PR #1, opened 2021-01-24
Add ability to define who can add proposals

ilblackdragon:


ilblackdragon:
FYI, going to rework this PR in the form of new permission model within a policy.

ilblackdragon:
This is in v2.

PR #8, opened 2021-03-07
Implementation of Sputnik v2

### LINKED ISSUE #4: v2 release feature set
"Hub" functionality on the factory contract:
 - [x] List of proposals and bounties as well
 - [ ] DAOs can pull from Hub and push proposals to Hub

New proposals types:
 - [x] Function call -- similar to multisig
 - [x] Poll: just voting on some question
 - [x] Transfer NEP-141 tokens
 - [x] New bounty
 - [x] Bounty XYZ is done
 - [x] Upgrade the Sputnik
 - [ ] Grace period change

New action types:
 - [x] Dismiss proposal -- instead of waiting for proposal to expire without voting, vote to dismiss it. This is different from rejecting which means that bond will be returned to the author.
 - [ ] Edit proposal -- allows to edit description and payout. This is useful when need to add things like discussion link or if the amount should be changed post discussion.
 - [ ] Push to "Hub" -- moves proposal into the factory
 - [ ] Pull from "Hub" -- moves proposal from the factory to this DAO

Upgradability:
 - [ ] New version of the contract goes into the factory
 - [x] Any of the specific instances can call upgrade proposal, which on passing requests new code from factory
 - [x] Supports migrating old data to new format.

Extend Policy:
 - [x] Add roles (e.g. "admin", "facilitator")
 - [x] Each role can have different threshold of voting for different proposal types and even different action types. Including defining who can do certain types of actions (#1)

This allows to have biggest membership list with different permissions, while still maintain smaller list of people who can make actions and decisions.

Token issuance:
 - [x] Set ticker, total balance and other parameters during DAO creation
 - [x] The initial amount is on the balance of the DAO itself (treasury) and needs be moved like others with proposals

One of the critical requirements is to make this code composable, so others can strip down features they don't need for their own version. E.g. token issuance is added as separate module and can be removed by removing few lines of code and trait implementation.

Note that all current DAOs will need to "exit" into new versions while keeping old names unusable.

### LINKED ISSUE #6: Implement bounties
Bounties is a way for a DAO to request work and show prepaid amount.

Bounty usually would contain next information:
 - `description`
 - `amount` --  amount to be paid out
 - `asset_name` -- accountId for asset to be paid (empty for native token?)
 - `repeat` -- how many times this bounty can be done
 - `max_deadline` -- maximum time from claim that can be spent on this bounty

Bounty life cycle:
 - Bounties can be added via proposal. When proposal will be voted in, this bounty will be added the bounty list.
 - Non claimed bounties can be removed via another proposal type.
 - Claim bounty - any one can call `bounty_claim(id, deadline)` to identify they want to work on it. Claim requires a bond that will be held for the deadline. Deadline must be less or equal than `max_deadline` set by bounty maker.
     - Bounty can be claimed up to `repeat` times at the same time.
     - Bounty claimer can give up `bounty_giveup(id)`, receiving back their bond. Within first `X` days - they receive 100% of their bond. After that, bond size gets linearly reduced to 0 at `deadline`.
     - Bounty claimer can indicate that bounty is done via `bounty_done(id, details)`, which marks bounty claim resolved and creates a proposal to payout the bounty reward. Bond is moved from claim to payout proposal. If proposal passes - the bond gets returned in full.
     - If claim hits `deadline`, the bond is withheld and another person can claim it.
     - Proposal by council can be sent to free bounty from (specific) claim as well.

All bounties must be accounted for in the DAO to make sure that DAO can't spent money below outstanding bounties.

Generally, any payout that would withdraw more money than is available should fail at the adding proposal time.

### LINKED ISSUE #7: Support FT
To add support for FT, Sputnik contracts must be implementing FT Receiver pattern to keep track of the received balance.

The received balance is then can be used internally to identify which tokens this DAO have and how much value.

E.g.
```
   /// Keeps track of balances per FT account.
   ft_balances: LookupMap<AccountId, Balance>,
```

Payout proposal then can select `asset_account_id`. Empty asset name is considered base token.

ilblackdragon:
This is SpuntikDAO v2:
 - has integrated fungible token that can be used for permissions
 - extended version of policy that allows to have groups of with different voting rights
 - upgradability (both for this contract and for others)
 - support for transferring fungible tokens
 - bounties

Fixes #4 
Fixes #6 
Fixes #7


sputnikdao2/src/proposals.rs
```diff
@@ -0,0 +1,430 @@
...
+    /// Add member to given role in the policy. This is short cut to updating the whole policy.
+    AddMemberToRole { member_id: AccountId, role: String },
+    /// Remove member to given role in the policy. This is short cut to updating the whole policy.
+    RemoveMemberFromRole { member_id: AccountId, role: String },
+    FunctionCall {
+        receiver_id: AccountId,
+        method_name: String,
+        args: Base64VecU8,
+        deposit: U128,
+        gas: Gas,
```
evgenykuzyakov, line 51: Do we use `Gas` in multisig or `U64`?

sputnikdao2/src/proposals.rs
```diff
@@ -0,0 +1,430 @@
...
+        receiver_id: AccountId,
+        method_name: String,
+        hash: Base64VecU8,
+    },
+    /// Transfers given amount of `token_id` from this DAO to `receiver_id`.
+    Transfer {
+        token_id: AccountId,
+        receiver_id: AccountId,
+        amount: U128,
+    },
```
evgenykuzyakov, line 66: Should this be `FtTransfer`?

sputnikdao2/src/proposals.rs
```diff
@@ -0,0 +1,430 @@
...
+    pub(crate) fn internal_payout(
+        &mut self,
+        token_id: &AccountId,
+        receiver_id: &AccountId,
+        amount: Balance,
+    ) -> PromiseOrValue<()> {
+        if token_id == &env::current_account_id() {
+            self.token
+                .internal_withdraw(&env::current_account_id(), amount);
+            self.token.internal_deposit(&receiver_id, amount);
```
evgenykuzyakov, line 186: Maybe use `internal_transfer`?

sputnikdao2/src/proposals.rs
```diff
@@ -0,0 +1,430 @@
...
+            PromiseOrValue::Value(())
+        } else if token_id == BASE_TOKEN {
+            Promise::new(receiver_id.clone()).transfer(amount).into()
+        } else {
+            ext_fungible_token::ft_transfer(
+                receiver_id.clone(),
+                U128(amount),
+                None,
```
evgenykuzyakov, line 194:
```suggestion
                Some("Best DAO transfer ever".to_string()),
```

sputnikdao2/src/proposals.rs
```diff
@@ -0,0 +1,430 @@
...
+            ProposalKind::UpgradeSelf { hash } => {
+                upgrade_self(&hash.0);
+                PromiseOrValue::Value(())
+            }
+            ProposalKind::UpgradeRemote {
+                receiver_id,
+                method_name,
+                hash,
+            } => {
+                let code = env::storage_read(&hash.0).expect("ERR_NO_CODE_STAGED");
```
evgenykuzyakov, line 254: Do we want to panic here?

sputnikdao2/src/proposals.rs
```diff
@@ -0,0 +1,430 @@
...
+                Promise::new(receiver_id.clone())
+                    .function_call(
+                        method_name.clone().into_bytes(),
+                        code,
+                        NO_DEPOSIT,
+                        env::prepaid_gas() - env::used_gas() - GAS_FOR_UPGRADE_REMOTE_PROMISE,
+                    )
+                    .into()
```
evgenykuzyakov, line 262: Wonder if we should replace with this with low level as well. Since passing code around might be expensive. But hopefully it doesn't clone it. But reading it into memory and back is probably expensive.

sputnikdao2/src/proposals.rs
```diff
@@ -0,0 +1,430 @@
...
+                deposit,
+                gas,
+            } => Promise::new(receiver_id.clone())
+                .function_call(
+                    method_name.clone().into_bytes(),
+                    args.clone().into(),
+                    deposit.0,
+                    *gas,
+                )
+                .into(),
```
evgenykuzyakov, line 244:
```suggestion
            } => {
              Promise::new(receiver_id.clone())
                  .function_call(
                      method_name.clone().into_bytes(),
                      args.clone().into(),
                      deposit.0,
                      *gas,
                  )
                  .into()
            }
```

sputnikdao2/src/proposals.rs
```diff
@@ -0,0 +1,430 @@
...
+                        NO_DEPOSIT,
+                        env::prepaid_gas() - env::used_gas() - GAS_FOR_UPGRADE_REMOTE_PROMISE,
+                    )
+                    .into()
+            }
+            ProposalKind::Transfer {
+                token_id,
+                receiver_id,
+                amount,
+            } => self.internal_payout(token_id, receiver_id, amount.0),
```
evgenykuzyakov, line 268:
```suggestion
            } => {
              self.internal_payout(token_id, receiver_id, amount.0)
            }
```

sputnikdao2/src/proposals.rs
```diff
@@ -0,0 +1,430 @@
...
+            }
+            ProposalKind::AddBounty { bounty } => {
+                self.internal_add_bounty(bounty.clone());
+                PromiseOrValue::Value(())
+            }
+            ProposalKind::BountyDone {
+                bounty_id,
+                receiver_id,
+            } => self.internal_execute_bounty_payout(*bounty_id, receiver_id, true),
+            ProposalKind::Vote => PromiseOrValue::Value(()),
```
evgenykuzyakov, line 287:
```suggestion
            } => {
              self.internal_execute_bounty_payout(*bounty_id, receiver_id, true)
            }
            ProposalKind::Vote => {
              PromiseOrValue::Value(())
            }
```

sputnikdao2/src/proposals.rs
```diff
@@ -0,0 +1,430 @@
...
+                &token_id,
+                ONE_YOCTO_NEAR,
+                GAS_FOR_FT_TRANSFER,
+            )
+            .into()
+        }
+    }
+
+    /// Executes given proposal and updates the contract's state.
+    fn internal_execute_proposal(&mut self, proposal: &Proposal) -> PromiseOrValue<()> {
```
evgenykuzyakov, line 204: Why not consume `Proposal` here, so you don't need to clone?

sputnikdao2/src/proposals.rs
```diff
@@ -0,0 +1,430 @@
...
+
+    /// Act on given proposal by id, if permissions allow.
+    pub fn act_proposal(&mut self, id: u64, action: Action) {
+        let mut proposal = self.data_mut().proposals.get(&id).expect("ERR_NO_PROPOSAL");
+        // Check permissions for given action.
+        assert!(
+            self.data().policy.to_policy().can_execute_action(
+                self.internal_user_info(),
+                &proposal.kind,
+                &Action::RemoveProposal
```
evgenykuzyakov, line 360: Yay, found a bug
```suggestion
                &action
```

sputnikdao2/src/proposals.rs
```diff
@@ -0,0 +1,430 @@
...
+                    .data()
+                    .policy
+                    .to_policy()
+                    .is_token_weighted(&proposal.kind)
+                {
+                    self.ft_balance_of(sender_id.clone().try_into().unwrap()).0
+                } else {
+                    1
+                };
+                proposal.update_votes(sender_id, Vote::from(action), amount);
```
evgenykuzyakov, line 388: I don't think `update_votes` checks that the vote didn't happen before, so you can vote multiple times.

sputnikdao2/src/proposals.rs
```diff
@@ -0,0 +1,430 @@
...
+    pub votes: HashMap<AccountId, Vote>,
+    /// Submission time (for voting period).
+    pub submission_time: WrappedTimestamp,
+}
+
+impl Proposal {
+    /// Update count and map of votes.
+    pub fn update_votes(&mut self, account_id: AccountId, vote: Vote, amount: Balance) {
+        self.vote_counts[vote.clone() as usize] += amount;
+        self.votes.insert(account_id, vote);
```
evgenykuzyakov, line 148: Probably want to check that you haven't voted.
```suggestion
        assert!(self.votes.insert(account_id, vote).is_none(), "ERR_ALREADY_VOTED");
```

sputnikdao2/src/lib.rs
```diff
@@ -0,0 +1,312 @@
...
+near_sdk::setup_alloc!();
+
+#[cfg(target_arch = "wasm32")]
+const BLOCKCHAIN_INTERFACE_NOT_SET_ERR: &str = "Blockchain interface not set.";
+
+/// Container for all contract data.
+#[derive(BorshSerialize, BorshDeserialize)]
+pub struct ContractData {
+    /// DAO configuration.
+    pub config: Config,
```
evgenykuzyakov, line 33: You may want to wrap config into `LazyOption` to avoid deserializing it every time. And then extract `bond` from `config` and store it separately (as a copy). Otherwise, FTMetadata fields such as icon can be pretty expensive to deserialize, imaging a 500 bytes icon, and also description + purpose
```suggestion
    pub config: LazyOption<Config>,
```

sputnikdao2/src/lib.rs
```diff
@@ -0,0 +1,312 @@
...
+    /// Voting and permissions policy.
+    pub policy: VersionedPolicy,
```
evgenykuzyakov, line 35: If policy is large, then you can do the same.

sputnikdao2/src/lib.rs
```diff
@@ -0,0 +1,312 @@
...
+
+#[near_bindgen]
+impl Contract {
+    #[init]
+    pub fn new(config: Config, policy: VersionedPolicy) -> Self {
+        assert!(!env::state_exists(), "ERR_CONTRACT_IS_INITIALIZED");
+        let mut this = Self {
+            token: FungibleToken::new(b"t".to_vec()),
+            data: VersionedContractData::Current(ContractData {
+                config,
```
evgenykuzyakov, line 79: With lazy option
```suggestion
                config: LazyOption::new(b"f".to_vec(), Some(config))
```

sputnikdao2/src/views.rs
```diff
@@ -0,0 +1,100 @@
...
+#[near_bindgen]
+impl Contract {
+    /// Returns semver of this contract.
+    pub fn version(&self) -> String {
+        env!("CARGO_PKG_VERSION").to_string()
+    }
+
+    /// Returns config of this contract.
+    pub fn get_config(&self) -> Config {
+        self.data().config.clone()
```
evgenykuzyakov, line 34: with LazyOption, it is 
```suggestion
        self.data().config.get().unwrap()
```

evgenykuzyakov (commented):
Just a few comments

ilblackdragon:
Addressed @evgenykuzyakov comments. Merging this to start testing more with real use cases.
Note, this is recommended for now to use with full access key (factory allows to pass one via creation function).

PR #11, opened 2021-05-03
External token as voting token

### LINKED ISSUE #10: Support NEAR's & ETH FT tokens as voting token
Currently DAO only supports voting with it's internal token for voting.
This is because it just uses internal balance without any cross contract calls to fetch state.

Even if we assume that we want to fetch the balance of the given user via cross contract call, this opens up potential issues that user just keep moving tokens from one account to another and vote with them.
To prevent that, tokens must be "frozen" in some way during the vote or alternatively if user moves their tokens - their vote gets nullified.

One option is that user must deposit the tokens into the DAO first to be able to vote (e.g. stake to participate in the governance).

What are other options to handle this?

ilblackdragon:
Fixes #10


oysterpack:
Here's an idea for the voting FT ... use **staked NEAR** as the voting token via **STAKE** tokens. Here's how it would work:

On the `ft_on_transfer()` the DAO looks up the STAKE NEAR value records it in a callback.

Thus, users are effectively voting with staked NEAR, but are still able to collect earnings while the STAKE tokens are locked / delegated for voting.

... my new STAKE pool supports all this functionality and ready to integrate on testnet - for more info, see my latest [tutorial][1]

A voting token is an excellent use case for staked NEAR and enhances the network's total value.

[1]: https://learn.figment.io/network-documentation/near/tutorials/1-project_overview/8-stake-pool-contract#oysterpack-smart-stake-benefits

ilblackdragon:
Agree, this is one of the options and allowing this option and many others is the reason for this PR.

sputnik-staking/src/lib.rs
```diff
@@ -0,0 +1,217 @@
...
+pub const GAS_FOR_REGISTER: Gas = 10_000_000_000_000;
+
+/// Amount of gas for undelegate action.
+pub const GAS_FOR_UNDELEGATE: Gas = 10_000_000_000_000;
+
+/// TODO: this should be in the near_standard_contracts
+#[ext_contract(ext_fungible_token)]
+pub trait FungibleToken {
+    fn ft_transfer(&mut self, receiver_id: AccountId, amount: U128, memo: Option<String>);
+}
```
evgenykuzyakov, line 36: It is `use near_contract_standards::fungible_token::core_impl::ext_fungible_token;`

PR #24, opened 2021-07-04
Add GitHub tests

mikedotexe:
Small changes to `test-token` and `sputnikdao` contract to get them to build properly.

Then use a project-level Cargo.toml with workspace members.

GitHub will build and test each project on Ubuntu and OS X on each push to a pull request.

(Windows cannot be added at this time because simulation tests use a dependency with uses `jemalloc-sys` which does not have this platform support.)

This one has a one-line redundancy with #23 

New project-level `Cargo.toml` means a new lockfile there, hence the unusually large number of lines added.


mikedotexe:
Looks like it's not running the GitHub tests here. Probably because this is a PR from a fork branch or something. You may see the GitHub tests running here: https://github.com/mikedotexe/sputnik-dao-contract/pull/1

PR #26, opened 2021-07-04
sputnikdao2: fix vote before full amount of Ⓝ available

### LINKED ISSUE #25: Transfer proposal on sputnikdao2 contract can cause odd UX
This probably applies to bounties as well, but will keep this issue focused on the Transfer proposal kind. Also, this applies to the original sputnikdao contract for the Payout type, but again, we'll focus on the newer contract here.

First, there's nothing stopping someone from creating a Transfer proposal that pays an amount of native NEAR Ⓝ that exceeds the available balance on the contract. We should probably not allow the creation of a proposal like this. This will encourage people to add a deposit to the call or else transfer Ⓝ before creating the proposal.

This doesn't, however, solve all the problems. Someone could create 10 Transfer proposals for 50 Ⓝ when the contract balance is 100 Ⓝ.

Second, when voting on a Transfer proposal, don't allow the vote logic to continue if the payout in Ⓝ cannot happen. If we don't add a more helpful error message/code, the error appears in Wallet with an `ExecutionError` message:

>Exceeded the account balance.

This is an unfriendly message to someone who is simply voting on a proposal and happens to be the last person to trigger the payout.

Third, a frontend for these proposal doesn't have a great way to determine if the vote buttons/interface should be enabled, and if they shouldn't be enabled, the reason why.

Steps to reproduce, by placing this in a shell file:

```sh
#!/bin/sh

# Change these to your account ids
export CONTRACT_ID=sputnikdao2.mike.testnet
export CONTRACT_PARENT=mike.testnet

# Redo account (if contract already exists)
near delete $CONTRACT_ID $CONTRACT_PARENT
near create-account $CONTRACT_ID --masterAccount $CONTRACT_PARENT

# Set up (change the wasmFile location)
near deploy $CONTRACT_ID --wasmFile ~/near/sputnik-dao-contract/sputnikdao2/res/sputnikdao2.wasm
export COUNCIL='["'$CONTRACT_ID'"]'
near call $CONTRACT_ID new '{"config": {"name": "genesis2", "purpose": "test", "metadata": ""}, "policy": '$COUNCIL'}' --accountId $CONTRACT_ID

# Set up a Transfer proposal for 3,000 NEAR which is more than the account has
# Note that it is allowed and perhaps shouldn't be
near call $CONTRACT_ID add_proposal '{"proposal": {"description": "test", "kind": {"Transfer": {"token_id": "", "receiver_id": "demo.testnet", "amount": "3000000000000000000000000000"}}}}' --accountId $CONTRACT_ID --amount 1

# A vote happens and a deeper error emerges that may confuse folks
near call $CONTRACT_ID act_proposal '{"id": 0, "action": "VoteApprove"}' --accountId $CONTRACT_ID
```

The last line of voting will result in this kind of output:

```sh
…
  type: 'FunctionCallError',
  context: undefined,
  index: 0,
  kind: { ExecutionError: 'Exceeded the account balance.' },
  transaction_outcome: {
…
```

mikedotexe:
Fixes #25 although it's possible we want to take a different approach.

- Doesn't allow someone to create a new Transfer proposal where the payout is in NEAR and exceeds the account balance of the contract
- Doesn't allow anyone to vote for a Transfer proposal where the payout is in NEAR and similarly exceeds the account balance (because they might be the final vote that attempts to trigger the payout)
- Adds a view method `proposal_voting_ready` where you give a parameter for the proposal ID and get a tuple back on whether the vote is ready and if not, why not (a warning code that can be used by the frontend to, for instance, disable the vote button)

Added a couple unit tests to check for the new errors.

Here is a helpful shell file to run or copy/paste showing how it works with the changes in this PR:

```sh
#!/bin/sh

# Change these to your account ids
export CONTRACT_ID=sputnikdao2.mike.testnet
export CONTRACT_PARENT=mike.testnet

# Redo account (if contract already exists)
near delete $CONTRACT_ID $CONTRACT_PARENT
near create-account $CONTRACT_ID --masterAccount $CONTRACT_PARENT

# Set up
near deploy $CONTRACT_ID --wasmFile ~/near/sputnik-dao-contract/sputnikdao2/res/sputnikdao2.wasm
export COUNCIL='["'$CONTRACT_ID'"]'
near call $CONTRACT_ID new '{"config": {"name": "genesis2", "purpose": "test", "metadata": ""}, "policy": '$COUNCIL'}' --accountId $CONTRACT_ID

# Add proposal for a Transfer kind that pays out 19 NEAR
near call $CONTRACT_ID add_proposal '{"proposal": {"description": "test", "kind": {"Transfer": {"token_id": "", "receiver_id": "demo.testnet", "amount": "19000000000000000000000000"}}}}' --accountId $CONTRACT_ID --amount 1

# Show the balance
near state $CONTRACT_ID

# Show that voting is ready
near view $CONTRACT_ID proposal_voting_ready '{"id": 0}'

# Transfer enough NEAR such that the Transfer proposal cannot complete successfully
near send $CONTRACT_ID $CONTRACT_PARENT 90

# Show that voting is not ready
near view $CONTRACT_ID proposal_voting_ready '{"id": 0}'

# Show error when a user tries to vote along with log
near call $CONTRACT_ID act_proposal '{"id": 0, "action": "VoteApprove"}' --accountId $CONTRACT_ID
```


sputnikdao2/src/proposals.rs
```diff
@@ -410,6 +410,9 @@ impl Contract {
                         ValidAccountId::try_from(token_id.clone()).is_ok(),
                         "ERR_TOKEN_ID_INVALID"
                     );
+                } else {
+                    // Don't create a Transfer proposal for Ⓝ that cannot be paid.
+                    assert!(env::account_balance() >= amount.0, "ERR_NOT_ENOUGH_BALANCE")
```
ilblackdragon, line 415: The full balance is not transferrable, because of storage and also potentially there are other money which are needed for other usages (like already committed in bounties). 

It's unclear how to facilitate correct check here for me actually. I was originally thinking of tracking "available" balance and subtract every time there is "commitment" to pay out but it's unclear when to commit the funds.

ilblackdragon:
We can't really check at the proposal creation time as it's possible that funds arrive later.
Also given the check at approval time - it will fail with the same mistake if there is not enough available funds as current asset.
Going to close this and instead focus on recovery from failed proposal completion

PR #38, opened 2021-09-02
Upgrade near-sdk dependency to use version 4.0.0-pre.2

roshkins:
This pull request updates the code to use the new near-sdk crate, and fixes breaking changes to the contract.


sputnikdao2/src/bounties.rs
```diff
@@ -246,6 +246,7 @@ mod tests {
     /// Adds a bounty, and tests it's full lifecycle.
     #[test]
     fn test_bounty_lifecycle() {
+        near_sdk::env::set_blockchain_interface(MockedBlockchain::default());
```
austinabell, line 249: you don't need this
```suggestion
```

sputnikdao2/src/views.rs
```diff
@@ -41,7 +41,7 @@ impl Contract {
 
     /// Returns staking contract if available. Otherwise returns empty.
     pub fn get_staking_contract(&self) -> AccountId {
-        self.staking_id.clone().unwrap_or_default()
+        self.staking_id.clone().unwrap()
```
austinabell, line 44: This will have different behaviour than what previously was there. I don't think the pattern that existed before was good, but if you want to match exactly you can do something like this:
```rust
    pub fn get_staking_contract(self) -> String {
        self.staking_id.map(String::from).unwrap_or_default()
```

sputnikdao2/tests/utils/mod.rs
```diff
@@ -133,6 +133,6 @@ pub fn vote(users: Vec<&UserAccount>, dao: &Contract, proposal_id: u64) {
     }
 }
 
-pub fn to_va(a: AccountId) -> ValidAccountId {
-    ValidAccountId::try_from(a).unwrap()
+pub fn to_va(a: AccountId) -> AccountId {
+    AccountId::try_from(a).unwrap()
 }
```
austinabell, line 138: optionally, can just remove this altogether, this is essentially a no-op. (pain to remove out so I get why not)

austinabell (approved):
Very well done! This approval is just based on a quick scan of the migration and didn't really look into the test-nft functionality (assume this is going to be used as a template nft in a future change?)

Edit: nvm test-nft was removed!

roshkins:
> Very well done! This approval is just based on a quick scan of the migration and didn't really look into the test-nft functionality (assume this is going to be used as a template nft in a future change?)
> 
> Edit: nvm test-nft was removed!

Yeah, I saw that the removal wasn't pushed, so I quickly did it.

sputnikdao2/tests/utils/mod.rs
roshkins: Removed!

sputnikdao2/src/views.rs
roshkins: switched!

sputnikdao2/src/bounties.rs
roshkins: Vanquished!

sputnikdao/src/lib.rs
```diff
@@ -420,10 +417,11 @@ mod tests {
             // vec![accounts(0).as_ref(), accounts(1).as_ref()],
     #[test]
     fn test_basics() {
+        near_sdk::env::set_blockchain_interface(MockedBlockchain::default());
```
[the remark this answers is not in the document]
roshkins: I'll remove it and remember for next time.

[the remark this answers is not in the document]
roshkins: I'll remove it and remember for next time.

roshkins:
Okay, just updated my branch addressing your critiques.

sputnikdao2/src/proposals.rs
```diff
@@ -79,19 +73,19 @@ pub enum ProposalKind {
     /// For `ft_transfer` and `ft_transfer_call` `memo` is the `description` of the proposal.
     Transfer {
         /// Can be "" for $NEAR or a valid account id.
-        token_id: AccountId,
-        receiver_id: ValidAccountId,
+        token_id: String,
```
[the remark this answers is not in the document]
austinabell: Note: the reason I was suggesting keeping this type is that you also need custom deserialization from borsh to be backwards compatible, and there is no annotation for swapping out cleanly with borsh afaik

This part is slightly less important, but changed as suggested [`658138b` (#38)](https://github.com/near-daos/sputnik-dao-contract/pull/38/commits/658138b41c2efdd9aa1db8962c02b920c436aa26) means that the dao contract code cannot be updated without a state migration.

sputnikdao/src/lib.rs
```diff
@@ -565,7 +562,7 @@ mod tests {
         testing_env!(VMContextBuilder::new().build());
```
austinabell, line 562: You can remove all lines that setup a default context like this if you'd like. It's just replacing the default with another default. This was only needed before 4.0.

roshkins: removed

ilblackdragon (approved):

PR #48, opened 2021-11-03
Quitting the Dao

swfsql:
Partial implementation of #41

There are open questions, such was:

- quitting is considered a direct action that a user can call, alongside add_proposal and act_proposal .
  - (before it was considered a proposal that didn't depend on approval nor on policies).
- quitting doesn't cleanup any data besides removing the user account from the roles.
  - (before it would remove empty roles if the quitting user was the last one) .

- this is still missing the policies questions, as raised on #41.


sputnikdao2/src/policy.rs
```diff
@@ -269,8 +275,21 @@ impl Policy {
...
 
-    /// Returns set of roles that this user is memeber of permissions for given user across all the roles it's member of.
-    fn get_user_roles(&self, user: UserInfo) -> HashMap<String, &HashSet<String>> {
+    /// Removes `member_id` from all roles.  
+    /// Returns `true` if the member was removed from at least one role.
+    pub fn remove_member_from_all_roles(&mut self, member_id: &AccountId) -> bool {
+        let mut removed = false;
+        for role in self.roles.iter_mut() {
+            if let RoleKind::Group(ref mut members) = role.kind {
+                removed |= members.remove(member_id);
```
mikedotexe, line 284: bitwise or?

sputnikdao2/tests/test_quit.rs
```diff
@@ -0,0 +1,246 @@
...
+                ExecutionStatus::SuccessValue(_bytes) => Ok(res.unwrap_json::<bool>()),
+                ExecutionStatus::Failure(err) => Err(err.to_string()),
+                _ => panic!("unexpected status"),
+            }
+        };
+
+    // initial check,
+    // when nobody has quit yet
+    let roles = dao_roles();
+    {
```
mikedotexe, line 133: Curious why this is in its own block here?

mikedotexe (commented):
Overall looking good, thanks. The simulation tests in `test_quit` were a bit confusing to read with a mixture of closures and functions, but I think we can work with that. Wouldn't hurt to have a bit more high-level comments peppered throughout tests. But also, thanks for the details comments above the functions, that's very helpful.

I'm going to open a pull request on something small that seems worth doing a different way.

sputnikdao2/src/policy.rs
swfsql: That's supposed to be a shortcut to `removed = removed | /*..*/;` 
But bools end up using boolean logic instead of bitwise operations

sputnikdao2/tests/test_quit.rs
swfsql: I'll remove it, thx!
It's a leftover from before because there were more operations but I ended-up using the `dao_roles_ref`

swfsql:
Thanks for the review Mike! I had placed the closures into functions, which I agree makes it easier to reason about the code!

I'm closing this issue in favor of https://github.com/near-daos/sputnik-dao-contract/pull/66, which contains updates from main and updated binaries

Issue #43, opened 2021-10-15
Ensure payments with fungible tokens can attach storage_deposit


mikedotexe:
![Screen Shot 2021-10-15 at 12 23 25 PM](https://user-images.githubusercontent.com/1042667/137541876-92161236-5fc1-431f-b3b3-f302f5ca3548.png)

My comment on this:

We should definitely fix that in the contract.

To have the smoothest experience while maintaining security for fungible tokens, the function would be have a Batch Action on the contract that does two function calls.
1. storage_deposit with the optional argument registration_only set to true
2. then transfer
If registration_only is set to true, it'll refund if the account is already paid for and not panic or block the transaction. Like this implementation:
https://github.com/near/near-sdk-rs/blob/95954cad746502aeb64e42a8bd5bbbb1ce0df8cc/near-contract-standards/src/fungible_token/storage_impl.rs#L45-L70

ctindogaru:
I'm taking a look into this now.

ctindogaru:
The issue has been addressed in https://github.com/near-daos/sputnik-dao-contract/pull/49.

However, it has been decided to solve this issue on Astro DAO side, instead of the Sputnik side, so that PR was closed.

PR #49, opened 2021-11-10
[2nd attempt] Register the account before making a transfer

ctindogaru:
It addresses https://github.com/near-daos/sputnik-dao-contract/issues/43.

Note that a change in `near-sdk` is also required for this change to work. See https://github.com/near/near-sdk-rs/pull/630.


ctindogaru:
@mikedotexe

mikedotexe (changes requested):
Thanks for the effort here. I think perhaps you may not have found the `storage_balance_of`
as mentioned in this comment:
https://github.com/near/near-sdk-rs/pull/629#pullrequestreview-804314025

I think broadly, we don't want to be changing the SDK to address this particular issue and have the tools we need to solve the issue.

ctindogaru:
> Thanks for the effort here. I think perhaps you may not have found the `storage_balance_of` as mentioned in this comment: [near/near-sdk-rs#629 (review)](https://github.com/near/near-sdk-rs/pull/629#pullrequestreview-804314025)
> 
> I think broadly, we don't want to be changing the SDK to address this particular issue and have the tools we need to solve the issue.

I don't know how could I've missed that. I thought for some reason, the method is internal and I cannot call it. Amazing suggestion, will apply that!

sputnikdao2/src/proposals.rs
```diff
@@ -2,9 +2,10 @@ use std::collections::HashMap;
 use std::convert::TryFrom;
 
 use near_contract_standards::fungible_token::core_impl::ext_fungible_token;
+use near_contract_standards::storage_management::StorageBalance;
```
mikedotexe, line 5: Thanks for the effort on the SDK side here, @constantindogaru https://github.com/near/near-sdk-rs/pull/630

Given some delays I anticipate in getting this change pushed to a prerelease or release for the SDK crate, can be please duplicate this struct in this project instead of using the `near_contract_standards` crate? Let's add a note in there about the PR for us in the future, like:

```rust
// Struct duplicated while pending: https://github.com/near/near-sdk-rs/pull/630
#[derive(BorshDeserialize, BorshSerialize, Serialize, Deserialize)]
#[serde(crate = "near_sdk::serde")]
pub struct StorageBalance {
    pub total: U128,
    pub available: U128,
}
```

(And then import any Borsh or serde we need to)

I hate to be hacky, but I think this is best to unblock us as I'm aware of issues the SDK team are dealing with that will likely delay this effort.

mikedotexe (changes requested):

swfsql:
Hey @ctindogaru I have added a PR changing some gas values, and also changing when some promises are returned, and with that the tests get working!  
I think that it's ideal to have fixed values for subtraction on Gas like I've made, ideally I think that env::used_gas() should be used (whenever possible)

https://github.com/ctindogaru/sputnik-dao-contract/pull/2

sputnikdao2/src/proposals.rs
```diff
@@ -217,41 +218,218 @@ impl From<ProposalInput> for Proposal {
...
+            "ERR_UNEXPECTED_CALLBACK_PROMISES"
+        );
+        match env::promise_result(0) {
+            PromiseResult::NotReady => {
+                Promise::new(proposer_account.clone()).transfer(attached_deposit.0);
+                panic!("storage_deposit Error: Received PromiseResult::NotReady")
+            }
+            PromiseResult::Failed => {
+                Promise::new(proposer_account.clone()).transfer(attached_deposit.0);
+                panic!("storage_deposit Error: Received PromiseResult::Failed")
```
swfsql, line 291: For this case and the `PromiseResult::NotReady` above, from some tests I made it appears that Promises that are being created in a function are discarded if a panic happens. So I think the panics should be avoided in there, and if a failure signal is necessary, perhaps a bool can indicate it?

sputnikdao2/src/proposals.rs
```diff
@@ -217,41 +218,218 @@ impl From<ProposalInput> for Proposal {
...
+            "ERR_UNEXPECTED_CALLBACK_PROMISES"
+        );
+        match env::promise_result(0) {
+            PromiseResult::NotReady => {
+                Promise::new(proposer_account.clone()).transfer(attached_deposit.0);
+                panic!("storage_balance_of Error: Received PromiseResult::NotReady")
+            }
+            PromiseResult::Failed => {
+                Promise::new(proposer_account.clone()).transfer(attached_deposit.0);
+                panic!("storage_balance_of Error: Received PromiseResult::Failed")
```
swfsql, line 328: (tagging another place, similar to another [comment](https://github.com/near-daos/sputnik-dao-contract/pull/49/files#r751128207))

starpause:
Via Astro Ecosystem meeting: It's preferable to handle this on the front end at time of proposal.

ctindogaru:
Ok, thanks for the update!

Will close this PR then.

PR #57, opened 2021-11-16
Alternative Names

swfsql:
After reading some discussions of users, I made a rename of `RolePermission`, `RoleKind` and `Action` (and added [more comments](https://github.com/near-daos/sputnik-dao-contract/pull/57/files#r750146636) into `RolePermission` in particular).

I don't expect this to get merged since it has no new features and yet breaks a lot of the wasm interface (function, data names).

The objective is to just showcase a different set of names for some structures which would work well with the comments that were added.


sputnikdao2/src/policy.rs
```diff
@@ -60,21 +63,106 @@ impl RoleKind {
...
-    /// Name of the role to display to the user.
-    pub name: String,
-    /// Kind of the role: defines which users this permissions apply.
-    pub kind: RoleKind,
-    /// Set of actions on which proposals that this role is allowed to execute.
-    /// <proposal_kind>:<action>
-    pub permissions: HashSet<String>,
-    /// For each proposal kind, defines voting policy.
-    pub vote_policy: HashMap<String, VotePolicy>,
+pub struct Council {
```
swfsql, line 101: (This structure got most of the comments)

ilblackdragon:
I don't really agree with Role to Council renaming - it's indeed Role as in ACL.
This role can be defined by list of people (council), token holding, some bridge contract or whatever else. Can be completely programmatic potentially.

swfsql:
@ilblackdragon I see, I didn't know about the ACL connection.  
I tried to rename to other things before getting to Council (RoleDecision, Judge - then I saw forum users using Council, so I thought it could make sense). The main point would be a possible connection to the act of "actually setting the proposal state", as if the Council were the actor itself.

This could be a wrong or not a good way of seeing the system, but I tried to expand this view on the Council struct documentation. I saw them as actors who are trying to make decisions on proposals, and they can analyze artifacts when trying to make a decision - one of those would be users votes.  
In this way, the decision-making process of the Council can be potentially programmatic, and the connection to the Roles in ACL would be on the Council's `permissions` field, which will decide if users opinions will be registered as artifacts (which will then be visible to the Councils) or ignored by the system.  
So another main difference on this view is that users never really "act" on proposals, they just register their "opinion" about it instead.

swfsql:
I think I'll be closing this as it's supposed to be an exploratory PR, but which won't really be developed nor depended upon. Thanks!

PR #58, opened 2021-11-22
adding new method to retrieve claims by bountyIds

karpovdmitryod:


sputnikdao2/src/lib.rs
```diff
@@ -59,7 +60,7 @@ pub struct Contract {
     /// Bounties map from ID to bounty information.
     pub bounties: LookupMap<u64, VersionedBounty>,
     /// Bounty claimers map per user. Allows quickly to query for each users their claims.
-    pub bounty_claimers: LookupMap<AccountId, Vec<BountyClaim>>,
+    pub bounty_claimers: UnorderedMap<AccountId, Vec<BountyClaim>>,
```
ilblackdragon, line 63: this is potentially a massive change that requires redoing this part of state of the contract.
also your implementation that goes over every single bountly_claim is very inefficient and won't fit usually in gas limits.
need to add an alternative index instead that will allow to fetch this data easily. Or just query the indexer for this data.

ctindogaru: I'm against using non near standard structures as well.

`LookupMap` is designed by near to be very efficient for read/writes with the drawback that it's not possible to iterate through it as long as you don't know the keys. So if it's absolutely necessary to have the `get_claims_by_bounty_id` method, one way around this is to have a vector which holds all the keys from `bounty_claimers` and make sure that this vector and `bounty_claimers` are always in sync. However, it will take up space and the sync will be difficult to mentain.

Maybe you can try @ilblackdragon's approach, which I don't fully understand to be honest.

Issue #47, opened 2021-11-02
Subtract storage cost from bond when returning


mikedotexe:
This issue is in regards to the SputnikDAO v2 contract.

Since proposals are not removed when completed, (accepted, rejected…) they take up storage space. This will eventually chip away at the storage on the DAO contract. On a medium-to-long timeframe, this can suddenly make certain DAOs inoperable until someone funds DAO with more NEAR. This is preventable by taking into consideration the amount of storage a proposal uses.

For example, James Waugh set up an AstroDAO for Portland, Oregon. When you go through the interface of Astro, it asks you to start a DAO with 5 Ⓝ, and this is how much we have in that DAO.

Here is a transaction where a person creates a proposal to a DAO:
https://explorer.testnet.near.org/transactions/4i8XGzg8g3TLiZVQPyy8MLjB7RrMJnWKs3aKDkTjq8Gc
This costs the bond amount (1 Ⓝ) plus the transaction fee of:
0.0007097988201294 Ⓝ

Let's pretend that this DAO explicitly does function calls, member addition/removal, and payouts in fungible tokens.

After the above proposal is accepted or rejected, the bond is returned to the proposer, but the storage of the proposal persists. If you compare before and after the proposal, you'll see that the DAO contract loses:
0.000665812311892 Ⓝ

This is a low amount but means after a few thousand proposals the DAO will run out of the default 5 Ⓝ that AstroDAO asked for in the creation wizard process. Then no more proposals can be added until someone figures out they need to add more Ⓝ.

**Solution**
For the various proposal types, come up with expected or worst-case-scenario storage fees, and subtract that from the bond when it's being returned. This is not web2; not everything is free.

swfsql:
I believe that issue #41 can be related to this as far as storage space cleanup is concerned.  

Given the results of the test that you made, the DAO is put in a "unstable" storage usage regarding their users. Should it always be "stable" towards it's users?  
What I mean by saying "stable toward the users", would be that all users could quit while asking for a total internal cleanup to get back their storage deposits, and the DAO would be able to return to exactly it's initial storage requirements, while returning everyone's funds.  
Another way of considering is to say that for being unstable, the DAO tends to lose some storage funds over time while it's being used, as pointed out by you - that would be some "negative" instability towards zero funds. Another option would be a positive instability towards (slowly) infinite funds - which can always be re-invested or shared among users at particular levels anyway.

[...]

PR #61, opened 2021-11-24
Allow failed payouts

ilblackdragon:
Creates new Proposal status: `Failed` which indicates that proposal needs to be retried via calling `act` with `Finalize` action when the reason of failure has been fixed. The reason for failure will need to be discovered from the indexer.


sputnikdao2/src/policy.rs
```diff
@@ -269,7 +269,7 @@ impl Policy {
         env::log_str(&format!("ERR_ROLE_NOT_FOUND:{}", role));
     }
 
-    /// Returns set of roles that this user is memeber of permissions for given user across all the roles it's member of.
+    /// Returns set of roles that this user is member of permissions for given user across all the roles it's member of.
```
mikedotexe, line 272: I actually just saved a stash fixing this typo, too, heheh

sputnikdao2/src/proposals.rs
```diff
@@ -341,9 +341,47 @@ impl Contract {
...
             } => self.internal_execute_bounty_payout(*bounty_id, &receiver_id.clone().into(), true),
             ProposalKind::Vote => PromiseOrValue::Value(()),
+        };
+        match result {
+            PromiseOrValue::Promise(promise) => promise
+                .then(ext_self::on_proposal_callback(
+                    proposal_id,
+                    env::current_account_id(),
+                    0,
+                    GAS_FOR_FT_TRANSFER,
```
mikedotexe, line 351: We could consider `GAS_FOR_PROPOSAL_CALLBACK` or something. Not urgent.
It does seem like this match block is how we simplify the work that was attempted in #32 👍🏼

sputnikdao2/src/proposals.rs
```diff
@@ -27,6 +27,8 @@ pub enum ProposalStatus {
     Expired,
     /// If proposal was moved to Hub or somewhere else.
     Moved,
+    /// If proposal has failed when finalizing. Allowed to re-finalize again to either expire or approved.
+    Failed,
```
mikedotexe, line 31: Good idea keeping this more generic, as mine was `PayoutFailed`. Part of me likes it being a specific failure, but this is probably the better path.

sputnikdao2/src/proposals.rs
```diff
@@ -250,20 +251,23 @@ impl Contract {
...
                     GAS_FOR_FT_TRANSFER,
                 )
-                .into()
             }
+            .into()
         }
     }
 
+    fn internal_return_bond(&self, policy: &Policy, proposal: &Proposal) -> Promise {
+        Promise::new(proposal.proposer.clone()).transfer(policy.proposal_bond.0)
```
mikedotexe, line 260: At some point I would like to subtract the storage cost associated with the proposal, but that's not for this PR and captured in #47

mikedotexe (approved):

Issue #59, opened 2021-11-23
Policy is too large for wallet interactions


ilblackdragon:
When policy is too large, it's not possible to update it via Web Wallet.
Specifically allowing for policy per group and per proposal type.

starpause:
Delta/Granular/Path based updates to policy will also address the concern of snapshotted policy state being overwritten (when there are multiple proposals updating policy & the order they're passed will impact final policy state).

mikedotexe:
I think we're going to need more detail here in terms of how to approach this. I don't want to set people off on the wrong track.
I will note this is a resource for how one might create a JavaScript file that can be run to change a policy:
https://hackmd.io/lMmmRVXASuGf9Xrx75Pmlw?view#Add-proposal-to-change-policy

So what is the course of action we want to see here?

mikedotexe:
I'll give my suggestions here.
Add proposal type:

### `ChangePolicyForRole`

```rust
ChangePolicyForRole {
  name: String,
  role_permission: RolePermission
}
```

### `ChangeDefaultVotePolicy`

```rust
ChangeDefaultVotePolicy {
  policy: VotePolicy
}
```

### `ChangePolicyItem`

```rust
enum PolicyItem {
  ProposalBond,
  ProposalPeriod,
  BountyBond,
  BountyForgivenessPeriod
}

ChangeDefaultVotePolicy {
  item: PolicyItem,
  value: U128
}
```

Note that for this one, there are two items (`proposal_period` and `bounty_forgiveness_period`) that use U64 instead of U128 and we'll be casting them as such.

ctindogaru:
> I'll give my suggestions here. Add proposal type:
> 
> ### `ChangePolicyForRole`
> ```rust
> ChangePolicyForRole {
>   name: String,
>   role_permission: RolePermission
> }
> ```
> 
> ### `ChangeDefaultVotePolicy`
> ```rust
> ChangeDefaultVotePolicy {
>   policy: VotePolicy
> }
> ```
> 
> ### `ChangePolicyItem`
> ```rust
> enum PolicyItem {
>   ProposalBond,
>   ProposalPeriod,
>   BountyBond,
>   BountyForgivenessPeriod
> }
> 
> ChangeDefaultVotePolicy {
>   item: PolicyItem,
>   value: U128
> }
> ```
> 
[...]

PR #62, opened 2021-11-24
Updating policy partially via JSON paths

ilblackdragon:
Example to just rename 0th role to `test`: `update_policy '{"updates": [{"path": "roles/0/name", "new_value": "\"test\"\"}]}'`

### DIFF
--- a/sputnikdao2/src/lib.rs
+++ b/sputnikdao2/src/lib.rs
@@ -217,6 +217,7 @@ mod tests {
     use crate::proposals::ProposalStatus;
 
     use super::*;
+    use crate::policy::PolicyUpdate;
 
     fn create_proposal(context: &mut VMContextBuilder, contract: &mut Contract) -> u64 {
         testing_env!(context.attached_deposit(to_yocto("1")).build());
@@ -371,5 +372,27 @@ mod tests {
                 policy: VersionedPolicy::Default(vec![]),
             },
         });
+    }
+
+    #[test]
+    fn test_update_policy() {
+        let mut context = VMContextBuilder::new();
+        testing_env!(context.predecessor_account_id(accounts(1)).build());
+        let mut contract = Contract::new(
+            Config::test_config(),
+            VersionedPolicy::Default(vec![accounts(1).into()]),
+        );
+        testing_env!(context.attached_deposit(to_yocto("1")).build());
+        let id = contract.add_proposal(ProposalInput {
+            description: "test".to_string(),
+            kind: ProposalKind::UpdatePolicy {
+                updates: vec![PolicyUpdate {
+                    path: "roles/0/name".to_string(),
+                    new_value: "\"test\"".to_string(),
+                }],
+            },
+        });
+        contract.act_proposal(id, Action::VoteApprove, None);
+        assert_eq!(contract.get_policy().roles[0].name, "test");
     }
 }
--- a/sputnikdao2/src/policy.rs
+++ b/sputnikdao2/src/policy.rs
@@ -355,7 +355,6 @@ impl Policy {
         roles: Vec<String>,
         total_supply: Balance,
     ) -> ProposalStatus {
-        env::log_str(&format!("{:?}", roles));
         assert!(
             matches!(
                 proposal.status,
--- a/sputnikdao2/src/policy.rs
+++ b/sputnikdao2/src/policy.rs
@@ -160,6 +160,18 @@ pub struct Policy {
     pub bounty_forgiveness_period: U64,
 }
 
+/// Partial update to the policy.
+#[derive(BorshSerialize, BorshDeserialize, Serialize, Deserialize, Clone)]
+#[cfg_attr(not(target_arch = "wasm32"), derive(Debug, PartialEq))]
+#[serde(crate = "near_sdk::serde")]
+pub struct PolicyUpdate {
+    /// Path in the policy to update using JSON format.
+    /// e.g. `roles/0/name`
+    pub path: String,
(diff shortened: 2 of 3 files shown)


sputnikdao2/src/lib.rs
```diff
@@ -372,4 +373,26 @@ mod tests {
             },
         });
     }
+
+    #[test]
+    fn test_update_policy() {
```
ctindogaru, line 378: This is a pretty sensible implementation that can lead to nasty hidden bugs. I would like to see more tests, in particular:

- updating `default_vote_policy`
- updating `proposal_bond` or `bounty_bond` which are U128
- updating `bounty_forgiveness_period` or `proposal_period` which are U64
- updating an entire role, not just the name. So the path is `roles/0` and `new_value` is an  `RolePermission` object serialized.

sputnikdao2/src/proposals.rs
```diff
@@ -91,6 +91,8 @@ pub enum ProposalKind {
     },
     /// Just a signaling vote, with no execution.
     Vote,
+    /// Update part of the policy.
```
ctindogaru, line 94: As stated on `AddMemberToRole` and `RemoveMemberFromRole`, can you add `This is short cut to updating the whole policy.` to the comment?

sputnikdao2/src/proposals.rs
```diff
@@ -91,6 +91,8 @@ pub enum ProposalKind {
...
+    UpdatePolicy { updates: Vec<PolicyUpdate> },
```
ctindogaru, line 95: From a naming point of view, `ChangePolicy` and `UpdatePolicy` types can lead to confusion for the end user, as they both seem to do the exact same thing.

Can we rephrase `UpdatePolicy` to avoid confusion? Some alternative names that pop into my mind now: `PartiallyUpdatePolicy` or `UpdatePolicyByJsonPaths` or `UpdatePolicyByRegex`.

sputnikdao2/src/proposals.rs
```diff
@@ -109,6 +111,7 @@ impl ProposalKind {
             ProposalKind::AddBounty { .. } => "add_bounty",
             ProposalKind::BountyDone { .. } => "bounty_done",
             ProposalKind::Vote => "vote",
+            ProposalKind::UpdatePolicy { .. } => "update_policy",
```
ctindogaru, line 114: Make sure you don't forget to also update the README file!

ctindogaru (commented):
Brilliant implementation, Illia. My only concern is that the end user may find it a bit difficult to use this new type.

Some examples in the readme on how to use it, would be amazing IMO.

```bash
near call $CONTRACT_ID add_proposal <TO_BE_COMPLETED> --accountId $CONTRACT_ID --amount 1
```
and showing how `TO_BE_COMPLETED` needs to look like for adding a new full role, changing the proposal bond or changing the weight kind for the default vote policy.

sputnikdao2/src/policy.rs
```diff
@@ -393,6 +404,47 @@ impl Policy {
...
+        // Walk the path to the item to change.
+        for i in 0..parts.len() {
+            let part = parts[i];
+            match item {
+                near_sdk::serde_json::Value::Array(subitems) => {
+                    let index = part
+                        .parse::<usize>()
+                        .expect(&format!("ERR_POLICY_PATH: {} must be integer", part));
+                    if i == parts.len() - 1 {
+                        subitems[index] = new_value_item.clone();
```
ctindogaru, line 425: what happens if the index exceeds the length of the `subitems` array?

sputnikdao2/src/policy.rs
```diff
@@ -393,6 +404,47 @@ impl Policy {
...
+                    } else {
+                        item = item
+                            .get_mut(index)
+                            .expect(&format!("ERR_POLICY_PATH: {} not found in policy", part));
+                    }
+                }
+                near_sdk::serde_json::Value::Object(subitems) => {
+                    if i == parts.len() - 1 {
+                        subitems.insert(part.to_string(), new_value_item.clone());
```
ctindogaru, line 434: what happens if the `part.to_string()` is a new key and not an existing one? Ex: `skadksdkas` instead of `proposal_bond`

sputnikdao2/src/policy.rs
```diff
@@ -393,6 +404,47 @@ impl Policy {
...
+    for update in updates {
+        let parts: Vec<&str> = update.path.split("/").collect();
+        let mut item = &mut policy_value;
+        let new_value_item =
+            near_sdk::serde_json::from_str::<near_sdk::serde_json::Value>(&update.new_value)
+                .expect("ERR_POLICY_VALUE");
+        // Walk the path to the item to change.
+        for i in 0..parts.len() {
+            let part = parts[i];
+            match item {
```
ctindogaru, line 419: Does this implementation work for adding a new group to the policy? As stated in https://hackmd.io/lMmmRVXASuGf9Xrx75Pmlw?view#Add-proposal-to-change-policy with the `influencers` group.

If not, I think it doesn't solve the original issue stated in https://github.com/near-daos/sputnik-dao-contract/issues/59

ctindogaru:
Hey Illia, to solve some of the issues mentioned in this PR, I've created https://github.com/near-daos/sputnik-dao-contract/pull/88.

ctindogaru:
Merged https://github.com/near-daos/sputnik-dao-contract/pull/88 instead. Although your implementation is brilliant, I find that #88 is easier to use for the end user and it also allows you to add/remove new groups from the policy.

PR #66, opened 2021-11-25
Quitting the DAO

swfsql:
Partial implementation of #41  
Has updated binaries. For changes, please see previous commits where the new binaries would not have been committed yet.

- Currently this doesn't make any changes to policies (nor any kind of cleanup) when a user quits.
- Quitting is considered a direct wasm function that a user can call, alongside `add_proposal` and `act_proposal`.


swfsql:
Closed because the main merge was split in two different merges (in the first one I branched-out during a merge, and that cause the merge to interrupt). Unifying into a single merge.

PR #67, opened 2021-11-25
Quitting the DAO

swfsql:
Partial implementation of #41  
Has updated binaries. For changes, please see previous commits where the new binaries would not have been committed yet.

- Currently this doesn't make any changes to policies (nor any kind of cleanup) when a user quits.
- Quitting is considered a direct wasm function that a user can call, alongside `add_proposal` and `act_proposal`.


sputnikdao2/tests/test_quit.rs
```diff
@@ -0,0 +1,415 @@
...
+
+/// Tests a role with Ratio = 1/2 with two members,
+/// when one member votes and then quits.  
+/// That single vote should not cause (nor allow) a state change
+/// in the proposal. That vote should be removed instead.
+///
+/// https://github.com/near-daos/sputnik-dao-contract/issues/41#issuecomment-971474598
+#[test]
+#[ignore]
+fn test_quit_removes_votes2() {
```
swfsql, line 362: This test requires the quitter's votes to be removed in order to pass, so it's currently ignored

sputnikdao2/src/basic_action.rs
```diff
@@ -0,0 +1,28 @@
...
+            env::panic_str("ERR_QUIT_WRONG_ACC");
+        }
+
+        let dao_name = self.get_config().name;
+        if dao_name != dao {
+            env::panic_str("ERR_QUIT_WRONG_DAO");
+        }
+
+        let mut new_policy = self.policy.get().unwrap().to_policy();
+        let removed = new_policy.remove_member_from_all_roles(&quitting_member);
```
TrevorJTClarke, line 23: @swfsql I think it should check that this DAO wouldn't lock, like if this was the last council / etc. Would be sad to have a user quit and lock the DAO for scenarios with diff groups or token weights that leave a community hanging.

sputnikdao2/src/lib.rs
```diff
@@ -8,11 +8,14 @@ use near_sdk::{
...
 use crate::bounties::{Bounty, BountyClaim, VersionedBounty};
-pub use crate::policy::{Policy, RoleKind, RolePermission, VersionedPolicy, VotePolicy};
+pub use crate::policy::{
+    Policy, ProposalPermission, RoleKind, RolePermission, VersionedPolicy, VotePolicy,
+};
 use crate::proposals::VersionedProposal;
 pub use crate::proposals::{Proposal, ProposalInput, ProposalKind, ProposalStatus};
 pub use crate::types::{Action, Config};
 
+mod basic_action;
```
TrevorJTClarke, line 18: Nit pick here - but can this file be called "council" or something, so its a file dedicated to council actions/abilities/controls?

sputnikdao2/tests/test_quit.rs
```diff
@@ -0,0 +1,415 @@
...
+            ("has_nobody", vec![]),
+            ("has_2", vec![]),
+            ("has_3", vec![&user3.account_id]),
+            ("has_23", vec![&user3.account_id]),
+            ("has_234", vec![&user3.account_id, &user4.account_id])
+        ]
+    );
+
+    // ok: user2 quits again
+    // (makes no change)
```
TrevorJTClarke, line 214: Shouldn't this error if user tries to quit that isnt council?

sputnikdao2/tests/test_quit.rs
```diff
@@ -0,0 +1,415 @@
...
+        vec![
+            ("all", vec![]),
+            ("council", vec![&root.account_id]),
+            ("has_nobody", vec![]),
+            ("has_2", vec![]),
+            ("has_3", vec![]),
+            ("has_23", vec![]),
+            ("has_234", vec![&user4.account_id])
+        ]
+    );
```
TrevorJTClarke, line 287: Add test for ensuring non-locking quit

TrevorJTClarke (changes requested):
@swfsql Woah - its been a while on this one, but really like the feature here. Its def something thats come up as a need, and is a step in the right direction on governance upgrades. Thank you for starting this!

PR #68, opened 2021-11-26
Fix bug where locked_amount and get_available_amount() show incorrect results

ctindogaru:
I've experienced this issue while working on a different PR.

`locked_amount` does not display the correct result. The amount of near locked for storage is higher than what `locked_amount` displays. Because of this, `self.get_available_amount()` which is using `locked_amount` also does not display the correct result.

If you try to call this from the contract:
`Promise::new(account_id).transfer(self.get_available_amount().0)` it will fail and it will tell you that if you transfer that amount the contract will not have the necessary NEAR to cover the storage costs.

### DIFF
--- a/sputnikdao2/src/lib.rs
+++ b/sputnikdao2/src/lib.rs
@@ -47,9 +47,6 @@ pub struct Contract {
     /// Voting and permissions policy.
     pub policy: LazyOption<VersionedPolicy>,
 
-    /// Amount of $NEAR locked for storage / bonds.
-    pub locked_amount: Balance,
-
     /// Vote staking contract id. That contract must have this account as owner.
     pub staking_id: Option<AccountId>,
     /// Delegated  token total amount.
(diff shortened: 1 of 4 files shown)


ctindogaru:
@mikedotexe

sputnikdao2/src/lib.rs
```diff
@@ -47,9 +47,6 @@ pub struct Contract {
     /// Voting and permissions policy.
     pub policy: LazyOption<VersionedPolicy>,
 
-    /// Amount of $NEAR locked for storage / bonds.
-    pub locked_amount: Balance,
```
ilblackdragon, line 51: this will break upgrade without having a specific migration code.

ilblackdragon (commented):
Your solution only accounts for the storage, but need to keep track of bonds as well, as they don't belong to the contract really.

Not sure what is the point of `get_locked_amount` otherwise - as it can retrieved directly from blockchain on how much storage is actually used without view call.

ctindogaru:
@ilblackdragon you're right, thanks for the good feedback! Hopefully, I've addressed all your concerns now :)

ilblackdragon (approved):
LGTM

PR #69, opened 2021-11-28
Factory manages contract versions

ilblackdragon:
Generalize factory to store multiple version of the contracts.
DAOs can just pull the specific version of the code instead requiring to upload it directly into their contract.

Adding ability for factory to "push" upgrade for a specific DAO if it is enabled.


sputnikdao-factory2/src/lib.rs
```diff
@@ -109,30 +80,115 @@ impl SputnikDAOFactory {
...
+
+    /// Tries to upgrade given account created by this factory to the latest code.
+    pub fn upgrade(&self, account_id: AccountId) {
+        self.assert_owner();
+        assert!(
+            self.daos.contains(&account_id),
+            "Must be contract created by factory"
+        );
+        self.factory_manager
+            .update_contract(account_id, self.get_latest_code_hash(), "update");
```
ctindogaru, line 92: Doesn't this call require the DAO to have an `update` method? I'm not seeing an implementation for `update` inside the DAO.

sputnikdao-factory2/src/lib.rs
```diff
@@ -1,106 +1,77 @@
...
-                0,
-                ON_CREATE_CALL_GAS,
-            ))
+        let callback_args = serde_json::to_vec(&json!({
+            "account_id": account_id,
+            "attached_deposit": U128(env::attached_deposit()),
+            "predecessor_account_id": env::predecessor_account_id()
+        }))
+        .expect("Failed to serialize");
+        self.factory_manager.create_contract(
```
ctindogaru, line 58: Compared with `update_contract`, this looks good since we have the `new` method inside the DAO.

ctindogaru (commented):
Amazing effort put into this PR 👍 . Well done.

I've finished my first round of review, but will take a second look.

sputnikdao-factory2/src/lib.rs
ctindogaru: You can ignore this comment, I've just seen you pushed a new commit with the implementation for `update`.

sputnikdao-factory2/src/lib.rs
ctindogaru: ignore this comment as well

sputnikdao-factory2/src/factory_manager.rs
```diff
@@ -0,0 +1,167 @@
...
+        callback_args: &[u8],
+    ) {
+        let code_hash: CryptoHash = code_hash.into();
+        let attached_deposit = env::attached_deposit();
+        let factory_account_id = env::current_account_id().as_bytes().to_vec();
+        let account_id = account_id.as_bytes().to_vec();
+        unsafe {
+            // Check that such contract exists.
+            assert_eq!(
+                sys::storage_has_key(code_hash.len() as _, code_hash.as_ptr() as _),
```
ctindogaru, line 127: In terms of readability, wouldn't be simpler to use the `env` wrappers? Ex:  `env::storage_has_key`, `storage_read` etc

sputnikdao2/src/upgrade.rs
```diff
@@ -0,0 +1,134 @@
...
+        // Update current contract with code from register 0.
+        let promise_id = sys::promise_batch_create(
+            current_id.as_bytes().len() as _,
+            current_id.as_bytes().as_ptr() as _,
+        );
+        // Deploy the contract code.
+        sys::promise_batch_action_deploy_contract(promise_id, u64::MAX as _, 0);
+        // Call promise to migrate the state.
+        // Batched together to fail upgrade if migration fails.
+        sys::promise_batch_action_function_call(
```
ctindogaru, line 74: I probably don't get this right, but it's not crystal clear to me why the `migrate` call is necessary here, since `migrate` will only read the contract code?

Is this to make sure that the `deploy` call succeeds? If that's the case, the `deploy` call can fail but the `migrate` call can succeed. Ex: if the contract code was already deployed and you're trying to deploy a newer version that fails, the `migrate` will succeed and return the old code.

Wouldn't be better to use `promise_batch_then` instead of `promise_batch_action_function_call` to make sure that deploy actually succeeded?

sputnikdao-factory2/src/factory_manager.rs
```diff
@@ -0,0 +1,167 @@
...
+const NO_DEPOSIT: Balance = 0;
+
+/// Factory manager that allows to store/load contracts by hash directly in the storage.
+/// Uses directly underlying host functions to not load any of the data into WASM memory.
+#[derive(BorshSerialize, BorshDeserialize)]
+pub struct FactoryManager {}
+
+impl FactoryManager {
+    /// Store contract from input.
+    pub fn store_contract(&self) {
```
ctindogaru, line 26: `store_contract` is a bit vague.

What about:
```
/// Store a new version of the DAO contract as WASM code.
/// The DAO contract will be saved as a key-value pair inside the storage.
/// The key is represented by the `sha256` representation of the WASM code.
/// The value is represented by the WASM code itself.
pub fn store_dao_contract_code(&self) {}
pub fn store_wasm_code_for_dao_contract(&self) {}
```

Insisting on specifying that the contract code is for the DAO, because at first, I thought it's the contract code for the factory itself.

sputnikdao-factory2/src/factory_manager.rs
```diff
@@ -0,0 +1,167 @@
...
+                0,
+                &NO_DEPOSIT as *const u128 as _,
+                (env::prepaid_gas() - env::used_gas() - GAS_UPDATE_LEFTOVER).0,
+            );
+            sys::promise_return(promise_id);
+        }
+    }
+
+    /// Create given contract with args and callback factory.
+    pub fn create_contract(
```
ctindogaru, line 111: `create_new_dao_contract`?

sputnikdao-factory2/src/factory_manager.rs
```diff
@@ -0,0 +1,167 @@
...
+            // Load the hash from storage.
+            sys::storage_read(code_hash.len() as _, code_hash.as_ptr() as _, 0);
+            // Return as value.
+            sys::value_return(u64::MAX as _, 0 as _);
+        }
+    }
+
+    /// Forces update on the given contract.
+    /// Contract must support update by factory for this via permission check.
+    pub fn update_contract(
```
ctindogaru, line 76: `update_existing_dao_contract`?

sputnikdao-factory2/src/lib.rs
```diff
@@ -109,30 +80,115 @@ impl SputnikDAOFactory {
             false
         }
     }
+
+    /// Tries to upgrade given account created by this factory to the latest code.
+    pub fn upgrade(&self, account_id: AccountId) {
```
ctindogaru, line 85: If not set otherwise, the factory itself will pay for all the upgrades right? And if we want to change that, we change the owner. Do I understand correctly?

sputnikdao-factory2/src/factory_manager.rs
ilblackdragon: nope, this is generic library - will move to near sdk later

sputnikdao-factory2/src/factory_manager.rs
ilblackdragon: nope, this is generic library - will move to near sdk later

sputnikdao-factory2/src/factory_manager.rs
ilblackdragon: nope, this is generic library - will move to near sdk later

sputnikdao-factory2/src/factory_manager.rs
```diff
@@ -0,0 +1,167 @@
...
+            sys::promise_return(promise_id);
+        }
+    }
+
+    /// Create given contract with args and callback factory.
+    pub fn create_contract(
+        &self,
+        code_hash: Base58CryptoHash,
+        account_id: AccountId,
+        new_method: &str,
```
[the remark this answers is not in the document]
ilblackdragon: nope, this is generic library - will move to near sdk later

sputnikdao-factory2/src/factory_manager.rs
```diff
@@ -0,0 +1,167 @@
...
+        }
+    }
+
+    /// Forces update on the given contract.
+    /// Contract must support update by factory for this via permission check.
+    pub fn update_contract(
+        &self,
+        account_id: AccountId,
+        code_hash: Base58CryptoHash,
+        method_name: &str,
```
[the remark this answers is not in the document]
ilblackdragon: this is generic library - will move to near sdk later

sputnikdao-factory2/src/factory_manager.rs
ilblackdragon: correct on `storage_has_key`, but `env::storage_read` doesn't allow to use registers.

sputnikdao-factory2/src/lib.rs
ilblackdragon: Generally, it is expected that storage is already covered before this is called (because this function is not payable) - given we are replacing contract in-place.
Whoever called the factory for the upgrade will be paying gas fees for this execution indeed.

sputnikdao2/src/upgrade.rs
ilblackdragon: current `migrate` is just a template for the actual migrate.
I've updated it to just be a better no-op.

Generally, when new version of the contract will be done - migrate should load the old state and save new state.

In case of migration function failing for whatever reason, if `promise_batch_then` was used - the contract will stay with the new contract code and old state. With single batch of two actions - if the second action fails, it will revert the first action (deploy of the contract) - recovering back to pre-upgrade state.

sputnikdao-factory2/src/lib.rs
```diff
@@ -109,30 +80,115 @@ impl SputnikDAOFactory {
...
+pub fn slice_to_hash(hash: &[u8]) -> Base58CryptoHash {
+    let mut result: CryptoHash = [0; 32];
+    result.copy_from_slice(&hash);
+    Base58CryptoHash::from(result)
+}
+
+/// Store new contract. Non serialized argument is the contract.
+/// Returns base58 of the hash of the contract.
+#[no_mangle]
+pub extern "C" fn store() {
```
ctindogaru, line 149: How are you supposed to call this method from the outside? It's not clear to me.

For example, let's say you're using `near cli`. How the call would look like?

ctindogaru: Just tried `near call FACTORY_ACCOUNT store {} --accountId FACTORY_ACCOUNT`. I don't know exactly why and how.

Also, the `store` method is reading the input from the context. How can you pass input to the context? In unit tests is easy to hack the input, but what about simulation tests/manual testing.

Is it possible to modify the input from the context?

ctindogaru: Is the input actually the arguments supplied to the call? In my example, it would be `{}`?

AngelBlock: > Is the input = the argument that is supplied to the call? In my example `{}`?

`
ARGS=`cat ../sputnikdao2/res/sputnikdao2.wasm  | base64`
near call FACTORY_ACCOUNT store '$ARGS' --base64 --accountId OWNER_ACCOUNT

`

AngelBlock: You need to bytes, not a json

ctindogaru: This is so so helpful. It's exactly what I needed to hear. Will give it a try, thank you very much for clarifying all of this!

I was struggling with getting the `wat` from the `wasm` and then passing it as an argument as a JSON or just simple string but none of them seemed to work.

AngelBlock: Wasn't clear to me too, at the start, thanks to Illia for the clarification.

sputnikdao2/src/upgrade.rs
```diff
@@ -0,0 +1,134 @@
...
+    assert!(
+        env::predecessor_account_id() == current_id
+            || (env::predecessor_account_id() == factory_info.factory_id
+                && factory_info.auto_update),
+        "{}",
+        ERR_MUST_BE_SELF_OR_FACTORY
+    );
+    unsafe {
+        // Load code into register 0 result from the promise.
+        sys::promise_result(0, 0);
```
ctindogaru, line 64: @ilblackdragon , I've just tested your entire code and everything runs smoothly except for the `upgrade` functionality.

I kept getting `Accessed invalid promise result index: 0` so I started debugging it. Apparently, the error gets generated from this line.

I've looked into it and something doesn't add up. You're trying to get the promise result from the previous call, but there is no call to get the result from. The only call is the one that you're actually in right now.

If you read the description of `promise_result` function, it says:
```
    /// If the current function is invoked by a callback we can access the execution results of the
    /// promises that caused the callback. This function returns the result in blob format and
    /// places it into the register.
```

The problem is that you don't call `promise_result` from the callback here. You're calling it from the call itself, so there's no result available.

My suggestion is to eliminate `promise_result` and get the contract code from the input context. Like you do on `store_contract`.

sputnikdao-factory2/src/factory_manager.rs
```diff
@@ -0,0 +1,160 @@
...
+            // Create a promise toward given account.
+            let promise_id =
+                sys::promise_batch_create(account_id.len() as _, account_id.as_ptr() as _);
+            // Call `update` method, which should also handle migrations.
+            sys::promise_batch_action_function_call(
+                promise_id,
+                method_name.len() as _,
+                method_name.as_ptr() as _,
+                u64::MAX as _,
+                0,
```
ctindogaru, line 95: I think what you actually need here is not 0, but `sys::value_return(u64::MAX as _, 0 as _);`. After this, on the DAO side, you can extract the value with `sys::input`.

sputnikdao2/src/upgrade.rs
ctindogaru:
```suggestion
        sys::input(0);
```

sputnikdao2/src/upgrade.rs
```diff
@@ -0,0 +1,134 @@
...
+    let current_id = env::current_account_id();
+    assert!(
+        env::predecessor_account_id() == current_id
+            || (env::predecessor_account_id() == factory_info.factory_id
+                && factory_info.auto_update),
+        "{}",
+        ERR_MUST_BE_SELF_OR_FACTORY
+    );
+    unsafe {
+        // Load code into register 0 result from the promise.
```
ctindogaru, line 63:
```suggestion
        // Load code into register 0.
```

ctindogaru, line 63:
```suggestion
        // Load code into register 0.
```

sputnikdao2/src/upgrade.rs
ctindogaru:
```suggestion
        sys::input(0);
```

ctindogaru:
Finished testing this PR in https://github.com/near-daos/sputnik-dao-contract/pull/76

Didn't find any issues apart from the one mentioned.

ctindogaru:
Had some discussions with Dmitry and we are having some concerns on the upgrade.

How it works now is that you store a key/value pair such that:
- the key is the sha256 hash of the value
- the value is the wasm file represented as a vector of bytes

Concerns:
- how do you know what contract code snapshot is represented by that wasm file? Since it's just an array of bytes, you have no way of telling what fixes/features it includes
- how do you know how many versions does the factory store and how do you know what to choose?

Suggestions:
- besides the vector of bytes, we should store some metadata that explains what the wasm file represents: snapshot of the contract code in the time (ex: commit id), version number: V1/V2/V3, description, github link or any other useful information.
- provide a way to get all the available versions so the users can pick which versions suits the best for them (ex: `get_contract_versions` will return `[v1: commit_id1, v1.1: commit_id2, v2: commit_id3, v2.1: commit_id4, v3: commit_id5`)

sputnikdao2/src/proposals.rs
```diff
@@ -91,6 +90,8 @@ pub enum ProposalKind {
     },
     /// Just a signaling vote, with no execution.
     Vote,
+    /// Change information about factory and auto update.
+    FactoryInfoUpdate { factory_info: FactoryInfo },
```
ctindogaru, line 94: please don't forget to include this in the readme

ilblackdragon:
Hm, it's surprising that everything worked given the contract storing method has assert in the opposite direction: https://github.com/near-daos/sputnik-dao-contract/pull/69/files#diff-3dd4449e06b690d9a171047ec7a41be850335de5eaa66aa192a71ac7880c1e24R155

ctindogaru:
> Hm, it's surprising that everything worked given the contract storing method has assert in the opposite direction: https://github.com/near-daos/sputnik-dao-contract/pull/69/files#diff-3dd4449e06b690d9a171047ec7a41be850335de5eaa66aa192a71ac7880c1e24R155

This is so nasty. It's my bad, it got away because I didn't attach any deposit to the call when testing: https://github.com/near-daos/sputnik-dao-contract/pull/76/files#diff-466303b3345b955ec8bba077825b0b9ead3fee8ce71d2959b32c80965048f116R52-R55

Good spot.

sputnikdao2/src/upgrade.rs
ilblackdragon: Good point. Specifically it should be sys::input if the call is from the factory and promise_result if it's a callback

ilblackdragon:
That's on me - I didn't have time to write tests for the factory which are really needed here.

If you have time this would really use good integration test for factory upgrade which would also have showcased the other issue you have found.

ilblackdragon:
RE: adding meta data -- that's a very good point. But a bit complicated in the case of storing contract, as that method doesn't parse arguments to be optimal on gas - just stores the whole blob of arguments into storage.

Potentially can add a separate method which allows to store metadata of the given code hash.

@ctindogaru how about merge this in and then you can add separate PR that would store metadata for the code hash?

ctindogaru:
Was thinking the same thing today. I don't want to make this PR bigger than it needs to be.

Since all the known issues have been resolved, will take another careful look tomorrow and probably merge it as it stands. Then, I will try to come up with 2 PRs: one for metadata, one for testing.

If we are implementing everything in this PR, it will get too complex and it will become harder to review it. Smaller increments over it is better.

sputnikdao-factory2/src/lib.rs
mikedotexe, line 149: Feels like this `store` function should also check for the owner, or else we might have random folks uploading blobs?
`self.assert_owner();` or something similar

sputnikdao2/src/proposals.rs
```diff
@@ -559,4 +565,30 @@ impl Contract {
...
+            .proposals
+            .get(&proposal_id)
+            .expect("ERR_NO_PROPOSAL")
+            .into();
+        assert_eq!(
+            env::promise_results_count(),
+            1,
+            "ERR_UNEXPECTED_CALLBACK_PROMISES"
+        );
+        let result = match env::promise_result(0) {
```
mikedotexe, line 585: Also, as a note we have great ergonomics around handling callbacks with results now. You can see an example here:
https://www.near-sdk.io/cross-contract/callbacks#whitelist-example
where we have essentially identical callbacks here:
https://github.com/mikedotexe/cross-contract-view/blob/06bdb3222622c824b9f2fe0a53536e6914435580/src/lib.rs#L56-L75

Except the second one uses `#[callback]` and saves quite a few lines.
cc @constantindogaru as an FYI 🎗️

mikedotexe: and actually looks like there might be further improvements according to the changelog:
https://github.com/near/near-sdk-rs/blob/master/CHANGELOG.md#400-pre3-10-12-2021

mikedotexe (approved):

ctindogaru (approved):
Approved and merged.

Will come up with small iterations over this.

sputnikdao2/src/proposals.rs
ctindogaru: Created https://github.com/near-daos/sputnik-dao-contract/issues/86 to address all of this.

PR #71, opened 2021-11-30
Returning more values from delegate/undelegate

ilblackdragon:
Returning more values from delegate/undelegate to allow for easier control from staking contract

Refers to https://github.com/ricktobacco/sputnik-dao-contract/commits


sputnikdao2/src/delegation.rs
```diff
@@ -31,12 +33,19 @@ impl Contract {
...
             .expect("ERR_NOT_REGISTERED");
-        self.delegations
-            .insert(account_id, &(prev_amount + amount.0));
+        let new_amount = prev_amount + amount.0;
+        self.delegations.insert(account_id, &new_amount);
         self.total_delegation_amount += amount.0;
+        (
+            U128(prev_amount),
+            U128(new_amount),
+            U128(self.total_delegation_amount),
```
ctindogaru, line 42: Can you use `self.delegation_total_supply()` instead of `U128(self.total_delegation_amount)`?

sputnikdao2/src/delegation.rs
```diff
@@ -45,8 +54,13 @@ impl Contract {
...
         assert!(prev_amount >= amount.0, "ERR_INVALID_STAKING_CONTRACT");
-        self.delegations
-            .insert(account_id, &(prev_amount - amount.0));
+        let new_amount = prev_amount - amount.0;
+        self.delegations.insert(account_id, &new_amount);
         self.total_delegation_amount -= amount.0;
+        (
+            U128(prev_amount),
+            U128(new_amount),
+            U128(self.total_delegation_amount),
```
ctindogaru, line 63: same as above

ctindogaru (approved):
Looks good to me 👍 . Only a minor suggestion.

Also, it's safe to merge it right away since the PR is not introducing any breaking changes, since `delegate` and `undelegate` have currently no return values.

PR #72, opened 2021-12-06
Fix ProposalKind types inside readme

ctindogaru:
Removes:
- ProposalKind::Mint
- ProposalKind::Burn
Adds:
- ProposalKind::SetStakingContract

It also removes the `=>` since it doesn't add any value and it's confusing what it's meant to be.


mikedotexe (approved):
I don't mind changing this. I did notice, for instance, that `FunctionCall` doesn't seem to point to any real method `call`

I'll post in Telegram and see if there are objections

ilblackdragon (approved):

ilblackdragon:
The reason for => is that those are names of the permissions. But we should explain that better.

PR #73, opened 2021-12-07
Actions on several proposals

fedencer:


sputnikdao2/src/proposals.rs
```diff
@@ -93,6 +93,14 @@ pub enum ProposalKind {
     Vote,
 }
 
+#[derive(BorshSerialize, BorshDeserialize, Serialize, Deserialize)]
+#[serde(crate = "near_sdk::serde")]
+pub struct ActionProposal {
```
ctindogaru, line 98: This sounds backwards to me. Can we rename it to `ProposalAction` instead?

sputnikdao2/src/proposals.rs
```diff
@@ -551,4 +559,18 @@ impl Contract {
...
         }
     }
+
+    /// Actions on the list of proposals
+    /// If the execution of the `act_proposal` method for any of the proposals causes panic,
+    /// the execution will be interrupted
+    pub fn act_proposal_multi(&mut self, actions: Vec<ActionProposal>) {
+        actions.into_iter().for_each(|proposal| {
+            env::log_str(&format!(
+                "Action {} by proposal {}",
```
ctindogaru, line 569: `Processing action {} for proposal {}`?

sputnikdao2/src/proposals.rs
```diff
@@ -551,4 +559,18 @@ impl Contract {
             log!("Memo: {}", memo);
         }
     }
+
+    /// Actions on the list of proposals
```
ctindogaru, line 563: `Act on given proposals, sequentially. The execution is interrupted and reverted if any of the proposal actions causes panic.`

ctindogaru (commented):
Thanks a lot for your contribution!

Minor suggestions.

AngelBlock:
> Thanks a lot for your contribution!
> 
> Minor suggestions.

Thanks for your input @ctindogaru , we get to it immediately.

ctindogaru (approved):
nice work 👍

sputnikdao2/src/proposals.rs
```diff
@@ -551,4 +559,16 @@ impl Contract {
             log!("Memo: {}", memo);
         }
     }
+
+    /// Act on given proposals, sequentially. The execution is interrupted and reverted if any of the proposal actions causes panic.
+    pub fn act_proposal_multi(&mut self, actions: Vec<ProposalAction>) {
```
ilblackdragon, line 564: how about naming it `act_proposal_batch`?

sputnikdao2/src/proposals.rs
```diff
@@ -551,4 +559,16 @@ impl Contract {
...
+        actions.into_iter().for_each(|proposal| {
+            env::log_str(&format!(
+                "Processing action {} for proposal {}",
+                proposal.action.to_policy_label(),
+                proposal.id
+            ));
+            self.act_proposal(proposal.id, proposal.action, proposal.memo);
```
ilblackdragon, line 571: there is an interesting issue here that act_proposal returns promises which may have callbacks attached as well. 
e.g. there potentially issues around callback trying to reverse state and failing due to complex interaction with other actions.

I think what make sense is to:
 -  add `finality: Option<bool>` to act_proposal, which if it's false, will not finalize the proposal automatically - so there is no execution involved.
 - from here call with `Some(false)` to not have any callback scheduled.
 - double check that finalize works for these proposals (right now I think it only works for expired and failed)

ilblackdragon (commented):
Adding security consideration

sputnikdao2/src/proposals.rs
AngelBlock: This was addressed here, can you have a quick look and comment please?

https://github.com/near-daos/sputnik-dao-contract/pull/73/commits/4c53e7bac55795873f068c813ff240fc3970d401

Thanks.

TrevorJTClarke:
@fedencer mind taking a minute to resolve latest conflicts?

sputnikdao2/src/proposals.rs
AngelBlock: @TrevorJTClarke has this solution to the issue been accepted?

ctindogaru:
Under what circumstances would you need batch actions on proposal?

ctindogaru:
I have a few concerns with this PR:
- it complicates the current logic by a lot
- it might introduce some nasty bugs: if someone has the last approval on a transfer proposal and sends a batch action with 3 approves on the proposal, that receiver can actually get funded 3 times.

AngelBlock:
> I have a few concerns with this PR:
> 
> * it complicates the current logic by a lot
> * it might introduce some nasty bugs: one example is: if someone has the last approval on a transfer proposal and sends a batch action with 3 approves on the same proposal, that receiver might actually get funded 3 times. This needs to be checked tho.

If this complicates the logic and opens to potential exploits, lets just cancel it altogether and I will be able to script it from frontend.

AngelBlock:
> Under what circumstances would you need batch actions on proposal?

There are DAOs which council need to vote for many proposals at the time, at the moment they use multiple devices to speed it up, for example 100 at the time - is quite challenging.

ctindogaru:
> > I have a few concerns with this PR:
> > 
> > * it complicates the current logic by a lot
> > * it might introduce some nasty bugs: one example is: if someone has the last approval on a transfer proposal and sends a batch action with 3 approves on the same proposal, that receiver might actually get funded 3 times. This needs to be checked tho.
> 
> If this complicates the logic and opens to potential exploits, lets just cancel it altogether and I will be able to script it from frontend.

If you can do that, it would be ideal @AngelBlock!

AngelBlock:
Parking of closing for reasons specified in the discussion thread.

PR #79, opened 2021-12-16
Remove duplicated validation on staking id

ctindogaru:
Remove unnecessary overhead. The validation already happens here: https://github.com/near-daos/sputnik-dao-contract/blob/main/sputnikdao2/src/proposals.rs#L442-L445


ctindogaru:
Closing this as I've just realised that if you're creating 2 `SetStakingContract` proposals simultaneously, approve the first one and then approve the second one, the second one will succeed although it should not since the staking id has already been set in the first one.

PR #87, opened 2022-01-12
Use V2 README as the main README for sputnik

ctindogaru:
Now that the V2 README has been improved (see https://github.com/near-daos/sputnik-dao-contract/pull/42), we can use it as a main one, since the current README is very outdated and it refers to v1.

Work done:
- move main README to the `sputnikdao` directory
- move README from `sputnikdao2` to the root directory


mikedotexe (approved):
This looks great, thank you

PR #88, opened 2022-01-13
Partially update the policy of a DAO by introducing smaller proposal kinds

ctindogaru:
Fixes https://github.com/near-daos/sputnik-dao-contract/issues/59.

**Problem:**
When the policy is being too large, using `ChangePolicy` proposal type is no longer feasible as the payload is too big and the request fails.

**Solution:**
Introduce 4 new proposal types to allow partially changing the policy:
- *ChangePolicyAddOrUpdateRole* - used to add a new role to the policy of the DAO. If the role already exists, update it.
- *ChangePolicyRemoveRole* - used to remove a role from the policy of the DAO.
- *ChangePolicyUpdateDefaultVotePolicy* - used to update the default vote policy from the policy of the DAO.
- *ChangePolicyUpdateParameters* - used to update the parameters from the policy of the DAO. Parameters include: proposal bond, proposal period, bounty bond, bounty forgiveness period.


sputnikdao2/src/policy.rs
```diff
@@ -237,8 +237,52 @@ impl VersionedPolicy {
...
-    ///
-    /// Doesn't fail, because will be used on the finalization of the proposal.
+    pub fn add_or_update_role(&mut self, role: &RolePermission) {
+        for i in 0..self.roles.len() {
+            if &self.roles[i].name == &role.name {
+                env::log_str(&format!(
+                    "Updating existing role in the policy:{}",
+                    &role.name
+                ));
+                let _ = std::mem::replace(&mut self.roles[i], role.clone());
```
mikedotexe, line 247: interesting approach. I know in the distant future we'd like to write smart contracts that don't use `std` at all, but think it's fine here :)

sputnikdao2/src/policy.rs
```diff
@@ -237,8 +237,52 @@ impl VersionedPolicy {
...
+                return;
+            }
+        }
+        env::log_str(&format!("Adding new role to the policy:{}", &role.name));
+        self.roles.push(role.clone());
+    }
+
+    pub fn remove_role(&mut self, role: &String) {
+        for i in 0..self.roles.len() {
```
mikedotexe, line 256: Could probably use `retain` here:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=1e84ddd396fd97ea7ab5d828e0de716a

sputnikdao2/src/proposals.rs
```diff
@@ -41,6 +41,17 @@ pub struct ActionCall {
     gas: U64,
 }
 
+/// Function call arguments.
+#[derive(BorshSerialize, BorshDeserialize, Serialize, Deserialize)]
+#[cfg_attr(not(target_arch = "wasm32"), derive(Clone, Debug))]
```
mikedotexe, line 46: don't think we need `Debug` here, but doesn't seem to affect the contract size

mikedotexe (approved):
Tested some of this including running this:

```bash
#!/bin/sh

# Change these to your account ids
export CONTRACT_ID=sputnikdao2.mike.testnet
export CONTRACT_PARENT=mike.testnet

# Redo account (if contract already exists)
near delete $CONTRACT_ID $CONTRACT_PARENT
near create-account $CONTRACT_ID --masterAccount $CONTRACT_PARENT

# Set up
near deploy $CONTRACT_ID --wasmFile ~/near/sputnik-dao-contract/sputnikdao2/res/sputnikdao2.wasm
export COUNCIL='["'$CONTRACT_ID'"]'
near call $CONTRACT_ID new '{"config": {"name": "genesis2", "purpose": "test", "metadata": ""}, "policy": '$COUNCIL'}' --accountId $CONTRACT_ID

near view $CONTRACT_ID get_policy

near call $CONTRACT_ID add_proposal '{
                                       "proposal": {
                                         "description": "update role",
                                         "kind": {
                                           "ChangePolicyAddOrUpdateRole": {
                                             "role": {
                                               "name": "council",
                                               "kind": {
                                                 "Group": [
                                                   "sputnikdao2.mike.testnet"
                                                 ]
                                               },
                                               "permissions": [
                                                 "*:Finalize",
                                                 "*:AddProposal",
                                                 "*:VoteApprove"
                                               ],
                                               "vote_policy": {}
                                             }
                                           }
                                         }
                                       }
                                     }' --accountId $CONTRACT_ID --amount 1


near call $CONTRACT_ID act_proposal '{"id": 0, "action": "VoteApprove"}' --accountId $CONTRACT_ID

near view $CONTRACT_ID get_policy
```

Thanks, looking good

sputnikdao2/src/proposals.rs
ctindogaru: Tried to mimic the `derive`s from `ProposalKind`, since `PolicyParameters` is used inside `ProposalKind`.

`ActionCall` (which is also used inside `ProposalKind`) also has the same `derive`s.

sputnikdao2/src/policy.rs
ctindogaru: haha nice one. Could use that, but then I would need to do some extra checking to verify if the role really existed in the first place so I can print some error otherwise.

sputnikdao2/src/policy.rs
ctindogaru: Interesting, I didn't know about that. Created an issue to try and address all the std use inside sputnik: https://github.com/near-daos/sputnik-dao-contract/issues/90

PR #94, opened 2022-01-24
Add contract metadata for the DAO when storing its contract code

ctindogaru:
The metadata will consist of:
- `code_hash`: sha256 hash of the contract code
- `version`: contract code version of the DAO (ex: V1, V2, V3, V4)
- `commit_id`: snapshot of the code from https://github.com/near-daos/sputnik-dao-contract. It represents the state of the code that generated the wasm contract code
- `readme`: link inside https://github.com/near-daos/sputnik-dao-contract that will point to a readme describing everything about the version

Tested manually via bash.

Usage:
```
// first store the contract code, which will return its hash
$ BYTES=$(cat sputnikdao2/res/sputnikdao2.wasm | base64)
$ near call sputnik.ctindogaru10.testnet store $BYTES --base64 --accountId sputnik.ctindogaru10.testnet --gas 100000000000000 --amount 10
// the above command will return 7ujcmhfQgQHnq7o7Bzz2VRGTSY2Yd97j8UmBStm9aEe4
// now store the metadata
$ near call sputnik.ctindogaru10.testnet store_contract_metadata '{"metadata": {"code_hash": "7ujcmhfQgQHnq7o7Bzz2VRGTSY2Yd97j8UmBStm9aEe4", "version": "v3", "commit_id": "tobecompleted", "readme": "readme_ceva"}}' --accountId sputnik.ctindogaru10.testnet
```

Get the contract information:
```
$ near view sputnik.ctindogaru10.testnet get_contract_metadata
```

Output example:
```
[
  {
    code_hash: 'GUMFKZP6kdLgy3NjKy1EAkn77AfZFLKkj96VAgjmHXeS',
    version: '',
    commit_id: '',
    readme: ''
  },
  {
    code_hash: '7ujcmhfQgQHnq7o7Bzz2VRGTSY2Yd97j8UmBStm9aEe4',
    version: 'v3',
    commit_id: 'tobecompleted',
    readme: 'readme_ceva'
  },
  {
    code_hash: '4ptDdxS1QweMYr3RZTTsw5RjfAwQaxPoRUkm5FwxvMMP',
    version: 'v4',
    commit_id: 'some_commit_id',
    readme: 'some_readme'
  }
]
```


ctindogaru:
@ilblackdragon this might interest you as well

sputnikdao-factory2/src/lib.rs
```diff
@@ -161,6 +203,7 @@ pub fn slice_to_hash(hash: &[u8]) -> Base58CryptoHash {
 pub extern "C" fn store() {
     env::setup_panic_hook();
     let contract: SputnikDAOFactory = env::state_read().expect("Contract is not initialized");
+    contract.assert_owner();
```
ctindogaru, line 206: This check will prevent random users to store blobs in the storage. Only the factory owner should be able to store new versions of the contract :)

sputnikdao-factory2/src/lib.rs
```diff
@@ -3,15 +3,31 @@ mod factory_manager;
...
-const LATEST_CODE_HASH_KEY: &[u8; 4] = b"CODE";
-const OWNER_KEY: &[u8; 5] = b"OWNER";
+#[derive(BorshSerialize, BorshDeserialize, Serialize, Deserialize)]
+#[cfg_attr(not(target_arch = "wasm32"), derive(Clone, Debug))]
+#[serde(crate = "near_sdk::serde")]
+pub struct DaoContractMetadata {
+    pub code_hash: Base58CryptoHash,
+    pub version: String,
+    pub commit_id: String,
+    pub readme: String,
```
mikedotexe, line 29: I don't think this is a big deal, but if we have the commit ID, you can see the README.md (or perhaps it's better if we created a markdown CHANGELOG.md like near-sdk-rs has with anchors to version changes: https://github.com/near/near-sdk-rs/blob/master/CHANGELOG.md#400-pre5-12-23-2021)

https://user-images.githubusercontent.com/1042667/151470608-626d6985-c0e1-41b1-96e8-4c469efc5a99.mp4

mikedotexe: So we may have enough info just given the commit it. But this is not strongly held as it's not that costly to have one extra field for `readme`

sputnikdao-factory2/src/lib.rs
```diff
@@ -24,31 +40,41 @@ pub struct SputnikDAOFactory {
 impl SputnikDAOFactory {
     #[init]
     pub fn new() -> Self {
+        assert!(!env::state_exists(), "The contract is already initialized");
```
mikedotexe, line 43: This line is curious to me…
Note that in this update: https://github.com/near/near-sdk-rs/blob/master/CHANGELOG.md#301-03-25-2021
We have:

>BREAKING #[init] now checks that the state is not initialized. This is expected behavior. To ignore state check you can call #[init(ignore_state)]

and I believe it happens here:
https://github.com/near/near-sdk-rs/blame/5dee86c9db5ca6ae30630d1f73fbe960f0e364b5/near-sdk-macros/src/core_impl/code_generator/impl_item_method_info.rs#L89-L91

But also, there are three places in the `near-sdk-macros` directory that have a similar check, but yours has the `!` in front of it, negating it. I'm not seeing that in the macros.

The two others:
https://github.com/near/near-sdk-rs/blob/5dee86c9db5ca6ae30630d1f73fbe960f0e364b5/near-sdk-macros/src/core_impl/code_generator/item_impl_info.rs#L497-L499 and 

and

https://github.com/near/near-sdk-rs/blob/5dee86c9db5ca6ae30630d1f73fbe960f0e364b5/near-sdk-macros/src/core_impl/code_generator/item_impl_info.rs#L579-L581

mikedotexe (changes requested):
Think I need a little clarity on the comments, but it's possible I'm misunderstanding

sputnikdao-factory2/src/lib.rs
```diff
@@ -124,22 +150,54 @@ impl SputnikDAOFactory {
...
 
+    pub fn store_contract_metadata(&self, metadata: DaoContractMetadata, set_default: bool) {
+        self.assert_owner();
+        let code_hash: CryptoHash = metadata.code_hash.into();
+        let storage_metadata = env::storage_read(CODE_METADATA_KEY);
+
+        if storage_metadata.is_none() {
+            let storage_metadata = vec![metadata];
+            let serialized_metadata =
+                BorshSerialize::try_to_vec(&storage_metadata).expect("INTERNAL_FAIL");
```
mikedotexe, line 177: This is less important for the factory, but as a reminder, `expect` actually bloats the smart contract a little, as it needs to capture information on line numbers and other info. So it's good to build the reflex of avoiding it for a slightly more strange syntax.

Instead of:

    let owner_id = self.owner_by_id.get(&token_id).expect("Token not found");

it's best to have:

```rust
fn expect_token_found<T>(option: Option<T>) -> T {
  option.unwrap_or_else(|| env::panic_str("Token not found"))
}
let owner_id = expect_token_found(self.owner_by_id.get(&token_id));  
```

As shown in our handy docs site:
https://www.near-sdk.io/reducing-contract-size/examples

sputnikdao-factory2/src/lib.rs
mikedotexe: I agree:

https://github.com/near-daos/sputnik-dao-contract/pull/69#pullrequestreview-849850581

🚀  Thank you

mikedotexe (approved):
Actually the item I had request changes for seemed to work contrary to my expectations. My other comments are hopefully helpful here

sputnikdao-factory2/src/lib.rs
ctindogaru: That information is very handy, but in this case I will chose readability over performance Mike.

That's only because `store_contract_metadata` will be called so rarely (maybe once every half a year?), that the performance increase is insignificant.

sputnikdao-factory2/src/lib.rs
ctindogaru: no, you're actually right. The check is no longer necessary since we have `init` macro. I've removed it, thanks for the suggestion!

sputnikdao-factory2/src/lib.rs
ctindogaru: `readme` was meant to be more like a changelog, than the current readme from the repo. But you're right, the name should reflect that. I've changed the name and also made the parameter optional. Thanks!

PR #97, opened 2022-01-27
Document the release process for upgrading the contract code for the DAOs

ctindogaru:


mikedotexe:
Thank you so much for this. I was just reviewing a pull request and thinking how handy it would be to flesh out a definitive guide like this. 🙌🏼

release_process.md
```diff
@@ -0,0 +1,324 @@
...
+
+***3. Init the factory:***
+```bash
+near call sputnik-factory.ctindogaru.testnet new '{}' --accountId sputnik-factory.ctindogaru.testnet --gas 100000000000000
+```
+
+***4. Download the current `wasm` code used for creating new DAOs:***
+
+```bash
+near view sputnikv2.testnet get_dao_list
```
ctindogaru, line 95: Need to remove this command since the list is so big that it fails. Will redirect people to the astro testnet instead.

TrevorJTClarke: Was gunna bring that up. I think if we do want to go this route, we have to do pagination. The other function should be the path devs use tho, since it already has pagination

TrevorJTClarke:
@ctindogaru docs look good - going to move them into a scripts folder that can be used as tools during the upgrade process. 
Also see: https://github.com/near-daos/sputnik-dao-contract/wiki/Sputnik-Factory-V2-Upgrade-Action-Plan

PR #99, opened 2022-02-01
Replace Vector with UnorderedMap and include version in DAO name

ctindogaru:


sputnikdao2/src/views.rs
```diff
@@ -1,9 +1,11 @@
+use near_sdk::borsh::{self, BorshDeserialize, BorshSerialize};
+
 use std::cmp::min;
 
 use crate::*;
 
 /// This is format of output via JSON for the proposal.
-#[derive(Serialize, Deserialize)]
+#[derive(BorshSerialize, BorshDeserialize, Serialize, Deserialize)]
```
ctindogaru, line 8: side-fix. This was currently failing when doing calls from the CLI.

PR #100, opened 2022-02-01
Added transfer batch proposal kind

okalenyk:


TrevorJTClarke:
Thanks @okalenyk will take a look and note findings.

okalenyk:
UPD: Added token 'msg' to TransferBatch proposal as well as the validation of the Proposal. Set max to 15 receivers since we can run out of gas otherwise.

ctindogaru:
Thank you a lot @okalenyk for your contribution 🤝 .

I'm trying to keep the core smart contract as simple as possible. So if any of the logic can be delegated to the user/UI, I would prefer to do so.

In this case, the `TransferBatch` proposal can be replicated by the user by having multiple `Transfer` proposals and then approving each one of them. I know it's a repetitive work, but I would prefer to not flood the contracts with batch logic, especially since the batch logic will be gas-limited to only 10-15 receivers.

I'm forced to postpone this PR for now until there is a real demand to have this sort of action.

PR #102, opened 2022-02-04
Proposal Deadline introduced

zavodil:
Please review how we can introduce time-based voting by setting `deadline` parameter for proposals. 
It means that proposal will be finalised after the deadline even without a majority vote. It may me useful for a really large DAOs. 
I don't know Sputnik DAO structure in details so may be my implementation is naive.

### DIFF
--- a/sputnikdao2/src/bounties.rs
+++ b/sputnikdao2/src/bounties.rs
@@ -1,7 +1,7 @@
 use near_sdk::borsh::{self, BorshDeserialize, BorshSerialize};
 use near_sdk::json_types::{U128, U64};
 use near_sdk::serde::{Deserialize, Serialize};
-use near_sdk::{env, near_bindgen, AccountId, Promise, PromiseOrValue};
+use near_sdk::{env, near_bindgen, AccountId, Promise, PromiseOrValue, Timestamp};
 
 use crate::*;
 
@@ -158,7 +158,7 @@ impl Contract {
     /// Only creator of the claim can call `done` on bounty that is still in progress.
     /// On expired, anyone can call it to free up the claim slot.
     #[payable]
-    pub fn bounty_done(&mut self, id: u64, account_id: Option<AccountId>, description: String) {
+    pub fn bounty_done(&mut self, id: u64, account_id: Option<AccountId>, description: String, deadline: Option<Timestamp>) {
         let sender_id = account_id.unwrap_or_else(|| env::predecessor_account_id());
         let (mut claims, claim_idx) = self.internal_get_claims(id, &sender_id);
         assert!(!claims[claim_idx].completed, "ERR_BOUNTY_CLAIM_COMPLETED");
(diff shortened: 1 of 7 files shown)


TrevorJTClarke:
@zavodil will need to review more in depth, but I believe this is actually duplicate logic. Maybe @constantindogaru has more insights, but the DAO policy configures a `proposal_period` that proposals utilize a check against.

See: https://github.com/near-daos/sputnik-dao-contract/blob/main/sputnikdao2/src/policy.rs#L156
and: https://github.com/near-daos/sputnik-dao-contract/blob/main/sputnikdao2/src/policy.rs#L397

I think a better approach would be to extend the use of this proposal_period, with constraints configured by DAO.

zavodil:
> @zavodil will need to review more in depth, but I believe this is actually duplicate logic. Maybe @constantindogaru has more insights, but the DAO policy configures a `proposal_period` that proposals utilize a check against.
> 
> See: https://github.com/near-daos/sputnik-dao-contract/blob/main/sputnikdao2/src/policy.rs#L156 and: https://github.com/near-daos/sputnik-dao-contract/blob/main/sputnikdao2/src/policy.rs#L397
> 
> I think a better approach would be to extend the use of this proposal_period, with constraints configured by DAO.

Yep I saw it, but current implementation of `proposal_period` is leading to proposal expiration only, my intension was to resolve proposal instead, that's why I introduced a new optional variable.

ctindogaru:
Hey @zavodil, I understand your point, but this shouldn't be needed. If the DAO is set up correctly, only certain members can vote on proposals and they should reach consensus on all of them. If the proposal hits the deadline and no consensus has been reached, it means there was a problem with the proposal itself or the DAO configuration (maybe more/less members should have the right to vote).

Now, assuming that we implement what you're suggesting, please review the following scenario:
- you have a group of 3 members that can vote on proposals => you need 2 approvals for a proposal to pass
- let's assume that the proposal deadline is set to 1 week
- let's assume that 2 members have taken a holiday for 2 weeks and they are away from their phone/laptop
- if we implement what you're saying, the 3rd member can take advantage of this and create the proposal while the other members are on holiday. He will create the proposal, wait for one week, approve it, the deadline will be reached and the proposal will be passed automatically.
- the other 2 members will return from holiday in one week and realise that the funds from a DAO have been transferred without their permission.

zavodil:
@ctindogaru Thank you for your feedback! Current DAO design is great for compact DAO (as you said in your example: 2-3 persons). My intention is to empower a really large DAO with hundreds of participants/councils, in which case it will be completely impossible to hit a quorum in this case and `deadline` proposals might come into play. Parameter `min_deadline_period' could be set in a DAO config, it could be longer than 3 days for DAO with huge budgets.

frol:
> My intention is to empower a really large DAO with hundreds of participants/councils, in which case it will be completely impossible to hit a quorum

Well, you can lower the quorum requirements to allow hitting it. Unconditional approval looks odd to me and still you need some trigger to execute the proposal, so in that case you may create a separate group with quorum of a single voter and grant it to some bot that will close the loops after some deadline, and this will happen outside of SputnikDAO logic.

I agree with the previous statements that this logic feels odd and complicates the configuration quite a bit, so I will close this old PR and we can revisit it in a new discussion thread

PR #104, opened 2022-02-07
group tests, so it's less buggy

deveusss:
Potentially fixing bug with tests ava


ctindogaru:
Thank you for the contribution, but I incline towards having more granularity inside the tests.

I'm in the favour of 7 smaller, isolated files than a giant one.

TrevorJTClarke:
> That was the exact bug with ava, with multiple files it consumes more memory, and bug with endpoints occurs...)

AVA concurrency could be at fault, however I agree with @ctindogaru with file structure remaining. Much better for maintainability long term. 
I think the real fix would be finding dependencies in mocks and splitting those into utilities on before/after methods to help with conflicts.
More test coverage is desired as well :)

deveusss:
> TrevorJTClarke
I think we will rewrite all the tests to rust version of workspaces, tests will run faster and we are more rusty)

TrevorJTClarke:
> tests will run faster and we are more rusty)

@deveusss that would be incredible! I def appreciate the move to Rust if you do in fact do so. I believe workspaces-rs to be ready but potentially less mature than workspaces-js FYI. Rust to test Rust is ideal for sure.

PR #105, opened 2022-02-07
Change ProposalKind::UpgradeSelf to call the factory for fetching the wasm code

ctindogaru:
DAOs no longer need to store the `wasm` code required for the upgrade inside their storage.

Instead, they will call the factory and get the `wasm` code from there.

By making the code only available in the factory:
- it will be easy to have a versioning system
- we will only have stable versions of the code that have been audited
- we free up a lot of storage space in the DAOs


ctindogaru:
testing not yet done. working on it

sputnikdao-factory2/src/lib.rs
```diff
@@ -126,15 +126,15 @@ impl SputnikDAOFactory {
         }
     }
 
-    /// Tries to upgrade given account created by this factory to the default code.
-    pub fn upgrade(&self, account_id: AccountId) {
+    /// Tries to update given account created by this factory to the specified code.
+    pub fn update(&self, account_id: AccountId, code_hash: Base58CryptoHash) {
```
TrevorJTClarke, line 130: Why the method name change here? I believe this to cause more issues than needed, esp for clients.

sputnikdao2/src/upgrade.rs
```diff
@@ -9,6 +10,7 @@ const DEFAULT_FACTORY_ID: &str = "sputnik-dao.near";
 const ERR_MUST_BE_SELF_OR_FACTORY: &str = "ERR_MUST_BE_SELF_OR_FACTORY";
 const SELF_MIGRATE_METHOD_NAME: &[u8; 7] = b"migrate";
 const UPDATE_GAS_LEFTOVER: Gas = Gas(5_000_000_000_000);
+const GAS_LEFTOVER: Gas = Gas(15_000_000_000_000);
```
TrevorJTClarke, line 13: Nit pick - Can you change to more concise name here? `FACTORY_UPDATE_GAS_LEFTOVER` perhaps?

TrevorJTClarke (changes requested):
@ctindogaru good to see the progress here!
A couple architectural thoughts/questions for you:
1. Following the model of the DAO submitting the code_hash, potentially the pitfall becomes a proposal that upgrades past where it can safely upgrade to (sans migrations, etc). I feel the DAO could accidentally skip or lie about the version it should upgrade to and potentially corrupt the DAO irreversibly. 
2. Versioning - Along the same thread as above, I believe the "V2", "V3" model may not work in our favor for long. 

Considering the above items, what are your thoughts on:
1. The factory keeps track of upgrade path, where the DAO just says "upgrade me" and the Factory has enough context to process the upgrade safely. `A -> B -> C`, rather than DAO saying: `A -> C`. I know it would also simplify the client UX as well.
2. To make above happen, Factory could implement semver, or simplified semver, so the factory could easily make decisions about next logical upgrade code_hash. It would also ensure no backward semi-unsafe downgraded code_hash.

I have been trying to formalize this in the document: https://github.com/near-daos/sputnik-dao-contract/wiki/WIP:-DAO-Factory-&-Registry
What are your thoughts on this path instead?

sputnikdao2/src/upgrade.rs
ctindogaru: sure, good suggestion

sputnikdao-factory2/src/lib.rs
ctindogaru: Because no one can stop you from downgrading a DAO to a previous version. If you provide the right hash, you can also downgrade.

ctindogaru:
> @ctindogaru good to see the progress here! A couple architectural thoughts/questions for you:
> 
> 1. Following the model of the DAO submitting the code_hash, potentially the pitfall becomes a proposal that upgrades past where it can safely upgrade to (sans migrations, etc). I feel the DAO could accidentally skip or lie about the version it should upgrade to and potentially corrupt the DAO irreversibly.
> 2. Versioning - Along the same thread as above, I believe the "V2", "V3" model may not work in our favor for long.
> 
> Considering the above items, what are your thoughts on:
> 
> 1. The factory keeps track of upgrade path, where the DAO just says "upgrade me" and the Factory has enough context to process the upgrade safely. `A -> B -> C`, rather than DAO saying: `A -> C`. I know it would also simplify the client UX as well.
> 2. To make above happen, Factory could implement semver, or simplified semver, so the factory could easily make decisions about next logical upgrade code_hash. It would also ensure no backward semi-unsafe downgraded code_hash.
> 
> I have been trying to formalize this in the document: https://github.com/near-daos/sputnik-dao-contract/wiki/WIP:-DAO-Factory-&-Registry What are your thoughts on this path instead?

Very good observations Trevor.

I completely agree with A -> B -> C approach and we should restrict any A -> C type of upgrades. This is because A -> B may need some data migration (that's automatically handle by the system), but B -> C does not need any data migration. So if you're doing A -> C, the data migration from A to B is skipped and your DAO may end up in an unpredictable state.

What it needs more thinking is:
- do we support downgrades? If we can handle data migrations from A to B, but from B to A, that's definitely not possible. So i'm inclined towards saying no here.
- where do we store the versioning for each DAO? In the factory? In the DAO? External database? Nowhere and is it handled by the UI?
- what versioning system are we using? counter? string? semver? The only problem that I have with using semver is the syntax, that may be very useful for certain applications, but doesn't seem very useful for sputnik: `major.minor.patch`.
1. `patch`: Since every release must be audited and heavily tested, we will never release patches (so the last part is useless as it will always be 0) because it's cost/time expensive.
2. `major`: This implies that there are some backward compatibility issues. Since the smart contracts, by default, cannot be upgraded if there are backwards compatibility issues and all of our contracts should be upgradable, I think this falls as well.
3. `minor`: this will be the only part from the `semver` that's actually going to be used. So why not just create a counter then?
4. Also, if we take the approach that the DAO can upgrade only one step at a time, we need to have a way of deciding what's the next version based on the current one. If we have `v2` then the next one is definitely `v3`. If we have `2` then the next one is definitely `3`. If we have `2.0.0`, how are you going to know which is the next version to upgrade to? Is it `2.0.1`? Or `2.1.0`? Or `3.0.0`?

Anther thought:
- a problem that comes up if we are taking an `upgrade me` approach instead of `update to specific hash` approach: The current method of upgrading (that is not documented at all) is through `UpgradeSelf { hash }` proposals. To keep everything simple for the user and also to easily track the upgrade history for a certain DAO, we should keep using the same proposal, but change the underlying implementation (this is what I've tried to achieve in this PR). So if the current state of art is that the proposal receives an hash and it upgrades to that hash, we should keep that logic, but maybe enforce a `one step at a time` approach on the UI side?

TrevorJTClarke:
> What it needs more thinking is:

TC: Adding inline thoughts to each

    do we support downgrades? We can handle data migrations from A to B, but B -> A is not possible. So i'm inclined towards saying no here.
TC: I really dont like the idea of downgrades. It introduces a lot of bad storage and potential attack surface. If a DAO needed a downgrade, we could suggest that they do a manual data migration to a DAO created by an earlier version. Which kinda brings up the idea of allowing creating a DAO at a specific version, which might be nice if there was demand and a solid reason? haha

    where do we store the versioning for each DAO? In the factory? In the DAO? External database? Nowhere and is it handled by the UI?
TC: I would like to explore the factory having more knowledge of the versions & DAO, it would require a little more cost for storage, but could allow for easier maintenance and improvements over time. The downside is more of a "centralized" point of failure, unless a DAO pointed to a different factory in the future. Since we're storing a list of all DAOs in the factory, could change to mapping instead. Could go both ways on this.

    what versioning system are we using? counter? string? semver? The only problem that I have with using semver is the syntax, that may be very useful for certain applications, but doesn't seem very useful for sputnik: major.minor.patch.
TC: Very good points on this - went down that path a little in my head, but liked your summary. We could likely keep to a single uint then, i just dont like the idea of String being the version. It could get messy to programmatically check version stance in the future. I like the thought of going `2` -> `3`, and dont allow `1`->`3` or `3`->`2`.

Anther thought:

    a problem that comes up if we are taking an upgrade me approach instead of update to specific hash approach: The current method of upgrading (that is not documented at all) is through UpgradeSelf { hash } proposals. To keep everything simple for the user and also to easily track the upgrade history for a certain DAO, we should keep using the same proposal, but change the underlying implementation (this is what I've tried to achieve in this PR). So if the current state of art is that the proposal receives an hash and it upgrades to that hash, we should keep that logic, but maybe enforce a one step at a time approach on the UI side?
TC: I see what you're saying, I guess since it hasn't been established yet as a solid standard, this is the 1 chance to change it or accept it. IF the proposal has to supply a hash, it will require a method for factory to help identify next hash, at which point the user could submit invalid data on purpose or otherwise. It would fail on factory side, but would be valuable lost time & fee. 
IF the proposal is generic `UpgradeSelf`, then the UI / Users can be out of the failure loop. We could still issue a helper view method to let client UIs show the next hash, so DAOs can confirm, but it would not require extra UX to create proposals.

----

## Decisions needed:
1. Can we agree upon the version? I'm suggesting a `u8` or similar, but not String
2. Can we agree upon how to handle the UpgradeSelf proposal? I'd like your feedback @ctindogaru  on above statement before making a decision.

ctindogaru:
> UnorderedMap<AccountId, u8> 

I hear you, but I find it very hard to keep everything in sync in the factory. I don't know how the migration would look like.

From v2->v3, the DAO does not use the factory to migrate. So if a DAO migrated through `UpgradeSelf { hash }` and it's now v3, when you call the factory to check for the DAO version, the factory will tell you that the DAO is v2. This is because the factory has no way of knowing that the DAO did the upgrade on its own...

So you can't rely on the factory to tell you the right versions, because the upgrades from v2 to v3 will happen outside of the factory.

> Can we agree upon the version? I'm suggesting a u8 or similar, but not String

u8 is perfect in my opinion and it's exactly what we need

> Can we agree upon how to handle the UpgradeSelf proposal? I'd like your feedback @ctindogaru on above statement before making a decision.

This is hard to change to exclude the hash, because in order for any v2 DAO to upgrade, the current code requires a hash. So even if we are changing the implementation, you still need to go through `UpgradeSelf { hash }` from v2 to v3. And if we decide to do things differently in v3, we need to introduce a new type of proposal `UpgradeMe`, because using `UpgradeSelf` without a hash will be an incompatible change.

ctindogaru:
> Can we agree upon the version? I'm suggesting a u8 or similar, but not String

Thinking better about this, I have fixed mixed feeling about it. u8 will limit you to a maximum of 256 versions. String provides more flexibility in that term + it allows you to specify a minor version of you want to.

Maybe you've fixed a bug that it's very urgent and you need to push it into production. You would call that version "v3.1" instead of "v4" because it hasn't been audited and you're mostly certain that it's safe to pursue it. I really like this kind of flexibility that the String provides....

sputnikdao2/src/upgrade.rs
```diff
@@ -1,14 +1,16 @@
 //! Logic to upgrade Sputnik contracts.
 
+use near_sdk::serde_json::json;
 use near_sdk::Gas;
 
 use crate::*;
 
 const FACTORY_KEY: &[u8; 7] = b"FACTORY";
-const DEFAULT_FACTORY_ID: &str = "sputnik-dao.near";
+const DEFAULT_FACTORY_ID: &str = "factory";
```
ctindogaru, line 9: we need to find a way to tell if we are on localnet/testnet/mainnet here

ctindogarus4f:
@TrevorJTClarke this is ready for review now

sputnikdao2/src/upgrade.rs
```diff
@@ -25,13 +26,30 @@ pub struct FactoryInfo {
     pub auto_update: bool,
 }
 
+pub fn get_default_factory_id() -> AccountId {
```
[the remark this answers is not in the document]
ctindogarus4f: Yes, on testnet this will work only with DAOs created from the official testnet factory. It can't know about the custom one that you've created.

sputnikdao2/src/upgrade.rs
TrevorJTClarke: Good idea - probably worth a separate issue

sputnikdao-factory2/src/lib.rs
TrevorJTClarke: I get that, but changing the method name is not good, is it absolutely necessary? I dont want clients supporting 2 method names for the same thing.

sputnikdao2/src/upgrade.rs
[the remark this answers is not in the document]
TrevorJTClarke: @ctindogaru i believe it could actually. Since every DAO is just a sub-account of the factory, the string would need to be split and then you have the factory. Thoughts?

TrevorJTClarke:
Noting discussion with @ctindogaru on version & UpgradeSelf:
1. Version will be implemented as `[u8; 2]` for simple semver
2. UpgradeSelf needs to continue including "hash" until it is possible for contracts to read `code_hash` for automatic upgrade checks.

sputnikdao-factory2/src/lib.rs
ctindogaru: Why do you need to support 2 method names? `upgrade` method was added very recently, it's not part of the deployed factory yet.

This is how the flow will go:
`dao::UpgradeSelf { hash } -> dao::upgrade_using_factory -> factory::update() -> dao::update()`

DAOs only need to know about `UpgradeSelf { hash }`, nothing more than that.

sputnikdao2/src/upgrade.rs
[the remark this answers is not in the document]
ctindogaru: I think now it should work with everything. Brilliant suggestion mister 👍

sputnikdao-factory2/src/lib.rs
TrevorJTClarke: Gotcha, didn't know this was a new method as well. In any case - update or upgrade is fine, i just want to be absolutely clear that downgrading is not supported, which can be implied with good naming.

TrevorJTClarke (approved):
@ctindogaru everything looking good! I will work toward scripts that match this flow for the upgrade docs

PR #109, opened 2022-02-09
Use simplified semver for DAO versioning instead of String

ctindogaru:


ctindogaru:
@TrevorJTClarke I know you can't wait to review this!

PR #111, opened 2022-02-14
Adding scripts to help the upgrade process

TrevorJTClarke:


scripts/deploy_factory_dev.sh
```diff
@@ -0,0 +1,229 @@
...
+  export RUSTFLAGS=''
+fi
+
+# TODO: Change to the official approved commit:
+COMMIT_V3=596f27a649c5df3310e945a37a41a957492c0322
+# git checkout $COMMIT_V3
+
+# build the things
+cargo build --all --target wasm32-unknown-unknown --release
+cp ../target/wasm32-unknown-unknown/release/*.wasm ../res/
```
ctindogaru, line 37: why not just call `./build` here?

scripts/deploy_factory_dev.sh
```diff
@@ -0,0 +1,229 @@
...
+
+if [ -d "res" ]; then
+  echo ""
+else
+  mkdir res
+fi
+
+cd "`dirname $0`"
+
+if [ -z "$KEEP_NAMES" ]; then
```
ctindogaru, line 25: I would get rid of this, it introduces unnecessary overhead.

scripts/deploy_factory_dev.sh
```diff
@@ -0,0 +1,229 @@
...
+# - submit DAO code v2 
+# - submit DAO metadata v2
+# - submit DAO code v3 
+# - submit DAO metadata v3
+# - upgrade DAO to v3 
+# - check if proposals return
+#### --------------------------------------------
+set -e
+
+if [ -d "res" ]; then
```
ctindogaru, line 17: I would get rid of this, it introduces unnecessary overhead.

scripts/deploy_factory_dev.sh
```diff
@@ -0,0 +1,229 @@
...
+export GAS_100_TGAS=100000000000000
+export GAS_150_TGAS=150000000000000
+
+# #### --------------------------------------------
+# #### Account & Data management for setup
+# #### --------------------------------------------
+# near call $FACTORY_ACCOUNT_ID delete_contract '{"code_hash":"Cn8QJe3SJk36b9NZky4SQuKtu9L2ktYYTCBdy2qKJ8Bq"}' --accountId $FACTORY_ACCOUNT_ID --gas $GAS_100_TGAS
+# near call $FACTORY_ACCOUNT_ID delete_contract '{"code_hash":"8RMeZ5cXDap6TENxaJKtigRYf3n139iHmTRe8ZUNey6N"}' --accountId $FACTORY_ACCOUNT_ID --gas $GAS_100_TGAS
+# near call $FACTORY_ACCOUNT_ID delete_contract '{"code_hash":"ZGdM2TFdQpcXrxPxvq25514EViyi9xBSboetDiB3Uiq"}' --accountId $FACTORY_ACCOUNT_ID --gas $GAS_100_TGAS
+# near call $FACTORY_ACCOUNT_ID delete_contract '{"code_hash":"6SQymHtmezR3u9zAtLBQdb8RWCXxwxnigqSH2mRTxecB"}' --accountId $FACTORY_ACCOUNT_ID --gas $GAS_100_TGAS
```
ctindogaru, line 62: why do we have 5 examples of the same thing?

scripts/deploy_factory_dev.sh
```diff
@@ -0,0 +1,229 @@
...
+# git checkout $COMMIT_V3
+
+# build the things
+cargo build --all --target wasm32-unknown-unknown --release
+cp ../target/wasm32-unknown-unknown/release/*.wasm ../res/
+
+export NEAR_ENV=testnet
+export FACTORY=testnet
+
+if [ -z ${NEAR_ACCT+x} ]; then
```
ctindogaru, line 42: i would remove this check as well.

scripts/deploy_factory_dev.sh
```diff
@@ -0,0 +1,229 @@
...
+  # export NEAR_ACCT=sputnikv2.$FACTORY
+  export NEAR_ACCT=sputnikpm.$FACTORY
+else
+  export NEAR_ACCT=$NEAR_ACCT
+fi
+
+export FACTORY_ACCOUNT_ID=factory2.$NEAR_ACCT
+# export DAO_ACCOUNT_ID=croncat.sputnikv2.$FACTORY
+export DAO_ACCOUNT_ID=sputnikdao-dev-1644971179.factory2.sputnikpm.testnet
```
ctindogaru, line 51: We should have only 3 constants here: wallet_id, factory_id, dao_id. Anything more than that is just overkill.

scripts/deploy_factory_dev.sh
```diff
@@ -0,0 +1,229 @@
...
+# TIMESTAMP=$(date +"%s")
+# DAO_NAME=sputnikdao-dev-v2-$TIMESTAMP
+# DAO_ARGS=`echo '{"config": {"name": "'$DAO_NAME'", "purpose": "Sputnik Dev v2 DAO '$TIMESTAMP'", "metadata":""}, "policy": '$COUNCIL'}' | base64`
+# near call $FACTORY_ACCOUNT_ID create "{\"name\": \"$DAO_NAME\", \"args\": \"$DAO_ARGS\"}" --accountId $FACTORY_ACCOUNT_ID --gas $GAS_150_TGAS --amount 10
+# DEMO_DAO_ACCOUNT=$DAO_NAME.$FACTORY_ACCOUNT_ID
+
+# # some sample payouts
+# near call $DEMO_DAO_ACCOUNT add_proposal '{"proposal": { "description": "Sample payment", "kind": { "Transfer": { "token_id": "", "receiver_id": "'$NEAR_ACCT'", "amount": "1000000000000000000000000" } } } }' --accountId $NEAR_ACCT --amount 1
+# near call $DEMO_DAO_ACCOUNT add_proposal '{"proposal": { "description": "Sample payment", "kind": { "Transfer": { "token_id": "", "receiver_id": "'$NEAR_ACCT'", "amount": "1000000000000000000000000" } } } }' --accountId $NEAR_ACCT --amount 1
+# near call $DEMO_DAO_ACCOUNT add_proposal '{"proposal": { "description": "Sample payment", "kind": { "Transfer": { "token_id": "", "receiver_id": "'$NEAR_ACCT'", "amount": "1000000000000000000000000" } } } }' --accountId $NEAR_ACCT --amount 1
```
ctindogaru, line 96: again, not sure why we are having the same command multiple times..

ctindogaru:
This is an amazing effort, but I feel like it adds a lot of complexity, the scripts are hard to grasp and more complicated than they should be.

Instead, I think it would be more useful to show commands examples: how to add a transfer proposal/how to create an account etc and let the user decide what flow they want to test.

TrevorJTClarke:
> This is an amazing effort, but I feel like it adds a lot of complexity, the scripts are hard to grasp and more complicated than they should be.
> 
> Instead, I think it would be more useful to show commands examples: how to add a transfer proposal/how to create an account etc and let the user decide what flow they want to test.

This PR is for setting up reproducibility for upgrades. The idea is to have:
* dev script: quick commands you, me & anyone (like auditors) else can quickly uncomment for spot testing
* testnet script: a stable, full script that goes from current state to fully upgraded state contracts, so there i no guessing or making mistakes in commands.
* mainnet script: a stable, full script that goes from current state to fully upgraded state contracts, same as testnet but even more careful attention to reproducible results.

Thanks for feedback on the scripts themselves, let's keep this setup going and move toward only using scripts for the upgrade process rather than comment/examples. It will help reproducible results & maintainability. Upgrade is scary, scripts will keep us sane :)

scripts/deploy_factory_dev.sh
TrevorJTClarke: could - just lazy i guess ;)

scripts/deploy_factory_dev.sh
TrevorJTClarke: These are actually 3 different on-chain versions deployed, my goal here is to catalog real contracts so I can triple check the process of upgrades for all scenarios. think of it as coverage for v1, v2, v2-recent & then local

scripts/deploy_factory_dev.sh
TrevorJTClarke: For dev script its not practical to slim down the constants. it is possible, however it limits being able to step through the process of upgrade. I will leave

scripts/deploy_factory_dev.sh
TrevorJTClarke: could setup a loop, just lazy and need 3 proposals

PR #115, opened 2022-02-18
Fix backward compatibility change introduced in PR #88

ctindogaru:
This backward compatibility issue was introduced in https://github.com/near-daos/sputnik-dao-contract/pull/88 by adding new elements in the middle of enum.

Instead, add them to the end to avoid backward compatibility issues.

Tested manually via upgrade.


TrevorJTClarke (approved):
This looks great! good catch!

PR #116, opened 2022-02-18
Fix backward compatibility issue introduced in #38 

ctindogaru:
See https://github.com/near-daos/sputnik-dao-contract/pull/38 for more context.

They've changed it in Transfer proposal type, but they forgot to change it in the bounty.


ctindogaru:
This should be replaced by https://github.com/near-daos/sputnik-dao-contract/pull/131

PR #129, opened 2022-02-26
Fix backwards compatibility issue on get_staking_contract

ctindogaru:
This was introduced in https://github.com/near-daos/sputnik-dao-contract/pull/38.

Previously, the method was returning `''` when there was no staking contract set up.
Now, the method is returning `null` which will cause backwards compatibility issues.

This PR aims to switch the logic back to how it was previously and return an empty string.


TrevorJTClarke:
Great catch - thanks!

PR #131, opened 2022-03-01
Refactor OldAccountId logic merged in PR #130

ctindogaru:
See https://github.com/near-daos/sputnik-dao-contract/pull/38 and https://github.com/near-daos/sputnik-dao-contract/pull/130 for more context.


sputnikdao2/src/types.rs
```diff
@@ -64,3 +67,10 @@ impl Action {
         format!("{:?}", self)
     }
 }
+
+pub fn convert_old_to_new_token(old_account_id: &OldAccountId) -> Option<AccountId> {
```
TrevorJTClarke, line 71: @ctindogaru thoughts about commenting a flag about this?
Would like to see more comments about why this was added, and details of when it can be removed.

sputnikdao2/src/types.rs
```diff
@@ -1,11 +1,14 @@
...
 use near_sdk::json_types::Base64VecU8;
 use near_sdk::serde::{Deserialize, Serialize};
-use near_sdk::{Balance, Gas};
+use near_sdk::{AccountId, Balance, Gas};
 
-/// Account ID used for $NEAR.
-pub const BASE_TOKEN: &str = "";
+/// Account ID used for $NEAR in near-sdk v3.
+/// Need to keep it around for backward compatibility.
+pub const OLD_BASE_TOKEN: &str = "";
```
TrevorJTClarke, line 8: @ctindogaru consider `V2_FALLBACK_TOKEN` or similar to denote when and what this is for, so future can be removed.

sputnikdao2/src/types.rs
```diff
@@ -1,11 +1,14 @@
...
 
+/// Account ID that represents a token in near-sdk v3.
+/// Need to keep it around for backward compatibility.
 pub type OldAccountId = String;
```
TrevorJTClarke, line 12: @ctindogaru consider `V2AccountId` or similar to denote when and what this is for, so future can be removed.

TrevorJTClarke (approved):
@ctindogaru Please add some comments & see naming convention suggestions.

I am approving, as this tested working, however I think hygiene is needed for future selfs

sputnikdao2/src/types.rs
ctindogarus4f: I think that's a very good suggestion. Will write a comment about it.

sputnikdao2/src/types.rs
ctindogaru: V2 might be confusing because this is a near sdk issue and people will not know if we're referring to the near sdk version or the sputnik version.

Also, the upgrade documentation advise to prefix everything with Old. See https://www.near-sdk.io/upgrading/production-basics

sputnikdao2/src/types.rs
ctindogaru: BASE_TOKEN is like a convention in Sputnik, used for $NEAR.

See also https://github.com/near-daos/sputnik-dao-contract/blob/main/sputnikdao2/tests/utils/mod.rs#L27-L29.

So to keep this consistent, I would stick with BASE_TOKEN :D

PR #133, opened 2022-03-02
Task 90/remove std dependencies

GentBinakuSQA:


GentBinakuSQA:
The unit test fail with the description:

thread 'tests::test_vote_twice' panicked at 'called `Result::unwrap()` on an `Err` value: HostError(GuestPanic { panic_msg: "Index out of bounds" })', /home/gentbinaku/near/near-sdk-rs/near-sdk/src/environment/mock/mocked_blockchain.rs:110:78

PR #134, opened 2022-03-08
DAO self upgrade via proposals

TrevorJTClarke:
This PR allows DAOs to upgrade via 3 proposals:
1. Store the official factory DAO contract on their DAO account
2. Upgrade the DAO contract
3. Remove the upgrade blob and retrieve storage deposit

**Requirements**

- [x] Store blob on self
- [x] Remove blob on self, get funds
- [x] Test coverage


sputnikdao-factory2/src/lib.rs
```diff
@@ -138,6 +144,109 @@ impl SputnikDAOFactory {
...
+
+        unsafe {
+            // Create a promise toward given account.
+            let promise_id =
+                sys::promise_batch_create(account_id.len() as _, account_id.as_ptr() as _);
+            sys::promise_batch_action_function_call(
+                promise_id,
+                method_name.len() as _,
+                method_name.as_ptr() as _,
+                DAO_CONTRACT_INITIAL_CODE.len() as _,
```
ctindogarus4f, line 180: This should be changed. The mainnet factory will not have `DAO_CONTRACT_INITIAL_CODE`.

We should fetch the code from the storage via a hash instead. The hash will be given as a parameter to the `store_contract_self` method.

sputnikdao-factory2/src/lib.rs
```diff
@@ -138,6 +144,109 @@ impl SputnikDAOFactory {
             .update_contract(account_id, code_hash, "update");
     }
 
+    /// Allows a DAO to store the official factory version as a blob, funded by the DAO wanting to upgrade
+    /// Required to successfully upgrade a DAO via proposals (proposal to store blob, proposal to upgrade from local blob)
+    /// Only intended for sputnik v2 DAO's created by sputnik factory
+    /// Payment is needed to cover storage costs for code blob size, paid by the DAO and returned upon blob removal
+    /// See docs for usage reference
```
ctindogarus4f, line 151: What docs are you referring to? I would remove this line, the description you've provided is crystal clear :D

sputnikdao-factory2/src/lib.rs
```diff
@@ -138,6 +144,109 @@ impl SputnikDAOFactory {
...
+                DAO_CONTRACT_INITIAL_CODE.as_ptr() as _,
+                &env::attached_deposit() as *const u128 as _,
+                (env::prepaid_gas() - env::used_gas() - GAS_STORE_CONTRACT_LEFTOVER).0,
+            );
+            sys::promise_return(promise_id);
+        }
+    }
+
+    /// Allows a DAO to remove the blob stored in its DAO storage, and reclaim the storage cost
+    /// See docs for usage reference
```
ctindogarus4f, line 190: same as above

sputnikdao-factory2/src/lib.rs
```diff
@@ -138,6 +144,109 @@ impl SputnikDAOFactory {
...
+    /// since it was the "owner" of the blob stored on the DAO.
+    /// Send this balance back to the DAO, since it was the original funder
+    #[private]
+    pub fn on_remove_contract_self(
+        &mut self,
+        account_id: AccountId,
+        factory_account_balance: U128,
+    ) -> bool {
+        if near_sdk::is_promise_success() {
+            let reimburse_amount = env::account_balance().saturating_sub(factory_account_balance.0);
```
ctindogarus4f, line 242: This is tricky and can lead to problems. What if someone made a huge payment to the factory right before this callback got executed? You would send back the DAO a huge chunk of money instead of just those 6 $NEAR.

I would limit the amount of attached deposit to exactly 6 NEAR on the `store_contract_self` method and send a refund of 6 $NEAR here.

ctindogarus4f (commented):
this will simplify our lives a lot. (more like DAO owners lives)

Thanks, Trevor! It just needs small adjustments.

sputnikdao-factory2/src/lib.rs
```diff
@@ -138,6 +144,109 @@ impl SputnikDAOFactory {
...
+        let callback_args = serde_json::to_vec(&json!({
+                "account_id": &account_id,
+                "factory_account_balance": U128::from(env::account_balance())
+            }))
+            .expect("Failed to serialize");
+
+        unsafe {
+            // Create a promise toward given account.
+            let promise_id =
+                sys::promise_batch_create(account_id.len() as _, account_id.as_ptr() as _);
```
ctindogarus4f, line 205: ffti: I would replace these `sys` calls with `env` ones. They are much cleaner.

Example: https://github.com/near-daos/sputnik-dao-contract/blob/main/sputnikdao2%2Fsrc%2Fupgrade.rs#L103-L118

sputnikdao-factory2/src/lib.rs
```diff
@@ -138,6 +144,109 @@ impl SputnikDAOFactory {
...
+            sys::promise_batch_action_function_call(
+                promise_id,
+                method_name.len() as _,
+                method_name.as_ptr() as _,
+                u64::MAX as _,
+                0,
```
ctindogarus4f, line 211: `remove_blob` accepts the hash for the blob. We need to provide the hash here.

TrevorJTClarke:
@ctindogarus4f thanks for the great suggestions - i have addressed them all. Need to figure out why workspace tests are having gas issues.

sputnikdao2/tests-ava/README.md
```diff
@@ -17,4 +17,8 @@ To run only one test file:
 To run only one test:
 
     npm run test -- -m "root sets*" # matches tests with titles starting with "root sets"
-    yarn test -m "root sets*"       # same thing using yarn instead of npm, see https://yarnpkg.com/
\ No newline at end of file
+    yarn test -m "root sets*"       # same thing using yarn instead of npm, see https://yarnpkg.com/
+
+If debugging:
+
+    NEAR_WORKSPACES_DEBUG=true npm run test "**/upgrade*"
```
ctindogaru, line 24: `"**/upgrade*"` is out of context. I think simply `NEAR_WORKSPACES_DEBUG=true npm run test` should be enough.

sputnikdao2/tests-ava/__tests__/upgrade.ava.ts
```diff
@@ -0,0 +1,163 @@
...
+		test.is(new_proposal.description, 'Store DAO upgrade contract code blob');
+		test.is(new_proposal.proposer, 'test.near')
+		test.is(new_proposal.status, 'InProgress')
+		test.truthy(new_proposal.kind.FunctionCall)
+		test.is(new_proposal.kind.FunctionCall.receiver_id, `${factory.accountId}`)
+
+		await root.call(dao, 'act_proposal', { id: 0, action: 'VoteApprove' }, { gas: tGas(300) })
+		let passed_proposal_0: any = await dao.view('get_proposal', { id: 0 })
+		test.log(passed_proposal_0);
+		test.is(passed_proposal_0.status, 'Approved')
```
ctindogaru, line 89: Hey Trevor. The reason why this whole flow is failing is because you are not creating the DAO via the factory in the first place.

The DAO that you're using in this test is stored inside `dao.test.near`.

The factory that you're using in this test is stored inside `factory.test.near`.

Of course the test is failing, because dao should have an account which looks more like `dao.factory.test.near`. Otherwise the `store_contract_self` method inside the factory will fail with the following error:

```
        assert_eq!(factory_id, env::current_account_id().as_str(), "Wrong factory");
```

because factory_id will be `test.near` and env::current_account_id() will be `factory.test.near`

sputnikdao-factory2/src/lib.rs
```diff
@@ -138,6 +143,108 @@ impl SputnikDAOFactory {
...
+
+    /// Upon blob remove, compute the balance (if any) that got paid to the factory,
+    /// since it was the "owner" of the blob stored on the DAO.
+    /// Send this balance back to the DAO, since it was the original funder
+    #[private]
+    pub fn on_remove_contract_self(&mut self, account_id: AccountId, code_hash: Base58CryptoHash) -> bool {
+        if near_sdk::is_promise_success() {
+            // Compute the actual storage cost for an accurate refund
+            let hash: CryptoHash = code_hash.into();
+            let dao_contract_code = env::storage_read(&hash).expect("CODE_HASH_NONEXIST");
```
ctindogaru, line 238: what if the `hash` is not in the storage anymore because it was removed by the factory? The user will not get refunded and the money will stay in the factory.

TrevorJTClarke: Yeah, i thought about that as well. I realize thats a possibility, but I would _hope_ standard practice of factory will not remove versions until all DAOs confirmed upgraded beyond. 
I just couldnt find a way yet that helps users retrieve funds in that case, and is safe.

ctindogaru (approved):
nice work man!

Feel free to merge this one in, once the CI completes.

sputnikdao-factory2/src/lib.rs
ctindogaru: Anyway, this entire flow is used only for v2 -> v3 upgrade. The idea that the V3 code will be wiped out from the factory is very unlikely so we should be fine.

Issue #121, opened 2022-02-25
TESTCOV: Sputnik Factory: Ownership


TrevorJTClarke:
Create test coverage for the following:

# Ownership
### Changing Owner
- [x] Can get current owner
- [x] Fails if trying to set owner from non-owner account
- [x] Owner can be a DAO account
- [x] Owner gets successfully updated

PR #146, opened 2022-03-23
TESTCOV: Sputnik Factory: Ownership

nninkovicSQA:
Draft PR for Sputnik Factory Ownership test coverage #121 

Tests will be written in Rust in a test module inside a respected file (lib.rs).

Recommendations are very welcome.


nninkovicSQA:
We are not compleatly sure that solution for [owner_can_be_a_dao_account] is correct.

sputnikdao-factory2/src/lib.rs
```diff
@@ -432,4 +433,73 @@ mod tests {
...
+    fn test_factory_can_get_current_owner() {
+        let mut context = VMContextBuilder::new();
+        testing_env!(context
+            .current_account_id(alice())
+            .predecessor_account_id(alice())
+            .attached_deposit(10)
+            .build());
+        let mut factory = SputnikDAOFactory::new();
+
+        factory.create(alice(), "{}".as_bytes().to_vec().into());
```
TrevorJTClarke, line 451: @nninkovicSQA  - this is a method to create a DAO. 
For testing the factory, `SputnikDAOFactory::new();` is the method used for a new factory, which will define the owner as the same account this factory code is deployed to until changed via `set_owner`.

sputnikdao-factory2/src/lib.rs
```diff
@@ -432,4 +433,73 @@ mod tests {
...
+    fn test_factory_fails_setting_owner_from_not_owner_account() {
+        let mut context = VMContextBuilder::new();
+        testing_env!(context
+            .current_account_id(alice())
+            .predecessor_account_id(carol())
+            .attached_deposit(10)
+            .build());
+        let mut factory = SputnikDAOFactory::new();
+
+        factory.create(alice(), "{}".as_bytes().to_vec().into());
```
TrevorJTClarke, line 467: unnecessary in this case

nninkovicSQA:
Changed as required @TrevorJTClarke

sputnikdao-factory2/src/lib.rs
```diff
@@ -432,4 +433,69 @@ mod tests {
...
+            .predecessor_account_id(bob())
+            .attached_deposit(20)
+            .build());
+        let mut factory = SputnikDAOFactory::new();
+
+        factory.create(bob(), "{}".as_bytes().to_vec().into());
+
+        factory.set_owner(alice());
+
+        assert_eq!(factory.get_owner(), alice())
```
TrevorJTClarke, line 482: This logic is wrong - it needs to use the DAO account in the `set_owner` method, then check that the DAO account is owner in `get_owner`.
Since you're using bob as the name, it will be something like `bob.sputnik-dao.near` as the DAO account. 
So make sure to change `set_owner` to be the actual full dao name

TrevorJTClarke (changes requested):
Looks like almost complete! Just need to change 1 test - see comment

sputnikdao-factory2/src/lib.rs
TrevorJTClarke: yes, all fixed - thanks!

TrevorJTClarke (approved):

PR #151, opened 2022-03-25
Remove 'unsafe' keyword from the smart contracts

ctindogaru:
Use `env` calls instead of `sys` to remove unsafe keyword and ensure memory safety.

The issue was reported by the Halborn team while doing an audit of the sputnik smart contracts.

Original description of the issue:
Rust code that uses the unsafe keyword is considered unsafe, since all of the memory safety guarantees of Rust are not enforced there. It means that the code might be prone to vulnerabilities that would’ve been prevented by the compiler, such as Buffer Overflow, Double free, Use After free, and more.


sputnikdao2/src/upgrade.rs
```diff
@@ -61,6 +60,7 @@ pub(crate) fn internal_set_factory_info(factory_info: &FactoryInfo) {
 #[no_mangle]
```
TrevorJTClarke, line 60: Can we remove these now?

sputnikdao2/src/upgrade.rs
```diff
@@ -118,51 +115,24 @@ pub(crate) fn upgrade_using_factory(code_hash: Base58CryptoHash) {
 }
 
 #[allow(dead_code)]
```
TrevorJTClarke, line 117: can this be removed?

sputnikdao2/src/upgrade.rs
```diff
@@ -118,51 +115,24 @@ pub(crate) fn upgrade_using_factory(code_hash: Base58CryptoHash) {
...
-        );
-    }
+    let input = env::storage_read(hash).expect("ERR_NO_HASH");
+    let promise_id = env::promise_batch_create(receiver_id);
+    let attached_gas = env::prepaid_gas() - env::used_gas() - GAS_FOR_UPGRADE_REMOTE_DEPLOY;
+    env::promise_batch_action_function_call(
+        promise_id,
+        method_name,
+        &input,
+        NO_DEPOSIT,
```
TrevorJTClarke, line 135: @ctindogaru i get that we're transitioning this, however - shouldn't this accept attached deposit? certain methods this calls require deposits

TrevorJTClarke: Basically, it should just fwd any/all deposits

sputnikdao2/tests-ava/__tests__/upgrade.ava.ts
```diff
@@ -49,11 +49,12 @@ workspaceWithFactory.test(
         const six_near = toYocto('6');
         const default_code_hash = await factory.view('get_default_code_hash');
 
-        let result = await root
-            .createTransaction('upgradedao.factory.test.near')
-            .functionCall('get_last_proposal_id', {}, { gas: tGas(300) })
-            .signAndSend();
-        let proposalId = result.parseResult<number>();
+        let proposalId: number = await root.call(
```
TrevorJTClarke, line 52: ahhhh, breath of fresh air :D

TrevorJTClarke (approved):
Approving, but check comments for minor things.
I was able to test without issue in testnet, including upgrades

ctindogarus4f:
Hey Trevor, thanks for the feedback. I would not change any logic in this PR. If you'd like we can open a new one and address those issues :D. This PR should be strictly about removing the `unsafe` keyword.

ctindogarus4f:
I've accidentally opened this PR against the audit branch. Will close it and open the same PR against `main`.

PR #153, opened 2022-03-30
Remove 'unsafe' keyword from the smart contracts

ctindogaru:
Use `env` calls instead of `sys` to remove unsafe keyword and ensure memory safety.

The issue was reported by the Halborn team while doing an audit of the sputnik smart contracts.

Original description of the issue:
Rust code that uses the unsafe keyword is considered unsafe, since all of the memory safety guarantees of Rust are not enforced there. It means that the code might be prone to vulnerabilities that would’ve been prevented by the compiler, such as Buffer Overflow, Double free, Use After free, and more.


ctindogaru:
The same PR as https://github.com/near-daos/sputnik-dao-contract/pull/151, but open against `main` not `audit-v3`.

PR #154, opened 2022-03-31
Remove linebreaks when Base64 encoding

roshkins:
`base64` adds linebreaks when encoding. use -w 0 to remove them. Throws JSON errors otherwise


TrevorJTClarke:
@roshkins are you on linux?
I fully feel your pain - adding this change however doesn't work for the people most using this scripting... :( 
Therefore, the solution here is to keep this as a referencable fix somewhere, or setup a second set of scripts for linux users.

The bigger problem is actually how the deploy fails for code, which ultimately means linux users cannot do the scripts until near-cli fixes/supports the PR: https://github.com/near/near-cli/pull/938

PR #155, opened 2022-04-04
Do not allow users to self delegate tokens inside the staking contract

ctindogaru:
Issue reported by Halborn during the audit.

Full description:
The delegate() function in sputnik-staking/lib.rs does not check if sender_id == account_id; as a result, the function caller can delegate votes to itself.

### DIFF
--- a/sputnik-staking/src/user.rs
+++ b/sputnik-staking/src/user.rs
@@ -184,6 +184,7 @@ impl Contract {
         delegate_id: AccountId,
         amount: Balance,
     ) {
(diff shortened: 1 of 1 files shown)


TrevorJTClarke:
@ctindogaru In fact this is not a bug - you should be able to stake your tokens, delegate to yourself, so you can vote with the weight you carry based on your staked tokens. 
I believe the audit didn't understand this use case.

ctindogaru:
Closing this as self-delegation is a normal process and should not be forbidden.

PR #156, opened 2022-04-04
Prevent delegation to non registered users

ctindogaru:
Issue reported by Halborn during the audit.

Full description:
The `internal_delegate()` function does not check if the delegated user is registered or not, it is possible to delegate to non-registered users.

It is seen that there is a comment “`/// The other user must be registered.`”


sputnik-staking/src/user.rs
```diff
@@ -185,6 +185,7 @@ impl Contract {
         amount: Balance,
     ) {
         let mut sender = self.internal_get_user(&sender_id);
+        assert!(self.users.contains_key(&delegate_id), "ERR_NOT_REGISTERED");
```
TrevorJTClarke, line 188: yay!

TrevorJTClarke (approved):
TYSM :D

PR #162, opened 2022-04-22
Testcov sputnik factory view functions

herolind-sqa:
Still work in progress


TrevorJTClarke:
@nninkovicSQA had a look at this PR real quick - the changes are not good, something changed formatting which doesnt allow checking a diff properly. 
Please be careful when doing large formatting changes (for example, stick with `cargo fmt`). 
I will have to keep this PR unmerged and unreviewed until the formatting changes are reverted.
Thank you!

PR #165, opened 2022-04-25
Add DAO data model architecture

ctindogaru:
This PR is a 1:1 copy of https://github.com/near-daos/sputnik-dao-contract/wiki/%5BDRAFT%5D-DAO-Data-Model-Architectures to allow easy feedback (via PR comments) on the original proposal.

### DIFF
--- a/dao_data_model.md
+++ b/dao_data_model.md
@@ -1,0 +1,227 @@
+> NOTE: The following is implementation examples for discussion on future models for DAOs
+
+## DAO Registry Example
+```rust
+// This is just extending the normal dao data, allowing registries to offload logic & state
+pub struct DAOContract {  
+  /// Governance modules are unowned math libraries that are only responsible for data computation. They follow a rigid definition governance spec/interface that the DAO can utilize within proposals.
+  pub governance_modules: UnorderedSet<AccountId>,
+  /// Extensions keep track of registry of DAO approved extensions (keys), where state requirements are wrapped in value - if specific data needs to remain within the DAO vs inside the extension. (Think of DAO preferences or pagination indexes)
+  pub extension_modules: UnorderedMap<AccountId, Vec<u8>>,
+  
+  /// NOTE: Policy may need changes to support diff groups control over governance modules/extensions
+}
+```
+
+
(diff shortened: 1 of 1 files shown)


ctindogaru:
Can we take a more granular approach and only focus on the governance module for now? There are too many big changes happening at the same time (governance + extension + user).

I would start a different thread on `extension_modules` after we agree on the `governance` module.

ctindogaru:
Can we take a more granular approach and only focus on the governance module for now? There are too many big changes happening at the same time (governance + extension + user).

I would start a different thread on `extension_modules` after we agree on the `governance` module.

ctindogaru:
What voting mechanisms can we have for a given proposal? I can only think of 3:
- equal voting rights (the current one)
- custom token voting. Each DAO can have a governance token and based on the amount that each member has, they have different voting rights.
- weighted voting. You assign different weights for each member when you're creating the council.

dao_data_model.md
```diff
@@ -0,0 +1,227 @@
...
+```
+
+
+## Governance Module Interface
+```rust
+// no state please, math only
+// no owners allowed, confirm no access keys before use
+pub trait Governance {
+  /// Implements a function that call tally a set of data, returning the collated value
+  pub fn compute(&self, ballots: Base64VecU8) -> PromiseOrValue<Outcome>;
```
ctindogaru, line 72: Agree with this interface, but Outcome can be replace by the existing `ProposalStatus`.

ctindogaru:
To conclude on this PR, I would strip out the extension module logic and put it in a different PR.

Also, I feel the governance module is more complicated than it should be. You only need the account id that holds the governance module to offload all the voting logic there.

Ex:

`near call <gov_module> compute '{"votes": {"account1.near": "yes", "account2.near": "yes", "account3.near": "no"}}' --acountId <caller_id>`

Then, the gov_module would know the weights for  `account1.near`, `account2.near`, `account3.near` and also the quorum so it will know how to compute the result based on those weights + quorum.

ctindogaru:
Note that I did not review the extension module. I feel that we should have a separate topic on that.

PR #166, opened 2022-04-25
Add USER data model architecture

ctindogaru:
This PR is a 1:1 copy of https://github.com/near-daos/sputnik-dao-contract/wiki/%5BDRAFT%5D-USER-Data-Model-Architectures to allow easy feedback (via PR comments) on the original proposal.

### DIFF
--- a/user_data_model.md
+++ b/user_data_model.md
@@ -1,0 +1,67 @@
+> NOTE: The following is implementation examples for discussion on future models for DAOs
+
+## User Data
+```rust
+pub struct ComposableUser {
+  pub attestations: LookupMap<AttestationUuid, Vec<Attestation>>,
+  pub attributions: LookupSet<Attribution>,
+  // User approved list of accounts they allow to submit attestations or attributions on this account.
+  // Default could be set with preferences, to either always allow new or always require approval
+  pub notaries: LookupSet<AccountId>,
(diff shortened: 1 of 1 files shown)


ctindogaru:
@TrevorJTClarke 

To be honest, I am not sure what's the use case for all of this. Do you have some context on why we created this wiki in the first place? Why do we need a structure for the user etc?

Everything seems overwhelming and it introduces a high complexity to the smart contracts...

user_data_model.md
```diff
@@ -0,0 +1,67 @@
...
+## User Data
+```rust
+pub struct ComposableUser {
+  pub attestations: LookupMap<AttestationUuid, Vec<Attestation>>,
+  pub attributions: LookupSet<Attribution>,
+  // User approved list of accounts they allow to submit attestations or attributions on this account.
+  // Default could be set with preferences, to either always allow new or always require approval
+  pub notaries: LookupSet<AccountId>,
+  // Allows user to control who & types of data allowed to be stored
+  pub preferences: UnorderedMap<Hash, Vec<u8>>,
```
ctindogaru, line 12: this might be just me, but these names are not very intuitive. What attestations should represent? What attributions should represent? Why do we need notaries.

Maybe I'm not understanding the purpose entirely, but everything seems over-engineered...

PR #167, opened 2022-04-26
Factory View Fuctions Tests WIP

nninkovicSQA:


nninkovicSQA:
@TrevorJTClarke can we have some feedback on this pr, it would be very helpful. Thanks.

PR #168, opened 2022-04-27
Replace assertion logic by macro

ctindogaru:
This has been suggested by Halborn during their audit.

No client impact expected.

See https://www.near-sdk.io/contract-interface/private-methods for documentation on the `private` macro.


TrevorJTClarke (approved):
@ctindogaru awesome - this is great

PR #189, opened 2022-08-15
TESTCOV env setup.

nninkovicSQA:
Set up simulation env using workspaces-rs for future simulation testing.


frol:
near-sdk-sim is deprecated. I’d recommend using the latest SDK with near-workspaces and  https://github.com/hknio/wasmcov

PR #193, opened 2022-11-30
feat: upgrade factory from DAO

agileurbanite:
# feat: upgrade factory from DAO

Allow factory to be upgradeable from a DAO act_proposal outcome.

Allows storing a factory then voting on upgrading it's hash to the new factory version while also storing factory metadata to keep track of factory changes.


agileurbanite:
Curious to get @ilblackdragon thoughts if we should add the concept of guardians with unique permissions as ref has: https://guide.ref.finance/developers/guardians

sputnikdao-factory2/src/lib.rs
```diff
@@ -342,6 +371,42 @@ impl SputnikDAOFactory {
         }
     }
 
+    pub fn store_factory_metadata(
```
ilblackdragon, line 374: what is the point in metadata if upgrade factory is called directly without storing contract code first?

agileurbanite: The point is to just have a on-chain record referencing code_hash, commit id, and changelog url.  If this is accomplished via the DAO storing the blob and then subsequently making the UpgradeRemote proposal then I can remove.  I just thought it would be better to have referenced in the factory itself.

As a side note, even store_contract_metadata technically doesn't do anything as this data can be encapsulated even within the proposal description itself.  I can remove for now and revisit in a separate standards doc on factory upgrades.

agileurbanite: Created #196 to review store_contract_metadata

agileurbanite:
@ilblackdragon made a note offline about the lack of migrate function in the factory, will need to update this PR to reflect that. Converting back to draft until ready for review

sputnikdao-factory2/src/lib.rs
agileurbanite: Resolved above issues by removing store_factory_metadata for now will discuss in detail on #196

agileurbanite:
Removed storing factory metadata and applied migrate method to be invoked during upgrade

PR #206, opened 2024-09-10
Merging set of fixes for upgrading smart contracts on Nightshade v2

referencedev:
This contains mix of fixes:
 - Old fixes 2 year ago
 - New fixes for Nightshade v2

Merging it all into main so we have the up to date version that is going to be used by Ref Finance and probably other DAOs.


think-in-universe:
To fix the build failure, install the latest `binaryen` version will make the "Wasm copy and strip" step success. 

Replace the binaryen version 105 with version 119 in `.github/workflows/build.yml`:

https://github.com/near-daos/sputnik-dao-contract/blob/2b19edb50e4542e1b8a6769023f2318c70255ee9/.github/workflows/build.yml#L91-L92

=>

```sh
           curl -L https://github.com/WebAssembly/binaryen/releases/download/version_119/binaryen-version_119-x86_64-linux.tar.gz | tar xzf - 
           echo "`pwd`/binaryen-version_119/bin" >> $GITHUB_PATH 
```

think-in-universe:
The test failure is caused by the `sputnikdao2-gasfix` project which depends on near-sdk "3.1.0" and near-sdk-sim "3.1.0" that probably leads to the test build failure. 

Is the `sputnikdao2-gasfix` project indeed needed? Could it be merged into `sputnikdao2` project?

Issue #214, opened 2025-02-27
Design and implement indexer-less indexing for `act_proposal` events


frol:
Currently, if I would like to find the transaction that executed the proposal (with the last `act_proposal` function call), I have to index the whole blockchain or use some APIs that indexed it for me. This makes DAO UIs dependent on indexers. Based on our experience with SocialDB and DevHub's contract, we can store one extra field in each proposal - "last touched block height" and as such I can easily find the most recent `act_proposal` transaction call and also recursively get the previous "last touched block height" by calling the function on the historical block height to reconstruct the full history of updates to the proposal.

Thus, I won't be dependent on indexers and RPC will be enough to extract the history. Yet, indexers / caching services should also be in place to speed things up, but they can be also a lot slimmer and stateless if they can use this same "last touched block height" mechanism that would be much faster to backfill or reconsile if the cached/indexed data is lost.

ori-near:
Hi @frol – I spoke with Thomas, who said he can pick up this ticket as part of his work for https://github.com/NEAR-DevHub/neardevhub-treasury-dashboard/issues/157. If that is okay, can you please assign him to this ticket?

frol:
@Tguntenaar please leave a comment here, so I can assign you to the issue (otherwise GitHub does not allow me to)

Tguntenaar:
Hi @frol !

Norfolks:
I will continue on this request.

frol:
@Tguntenaar Just FYI, I am onboarding @Norfolks into the SputnikDAO indexer epic, so he will start by picking up your draft PR for this issue.

Norfolks:
After extensive discussions with @frol, we decided to make important changes to this PR and the Epic implementations.

- A proposal maintains a list of the last 20 blocks in which it was modified.
- The contract now has a global list of the last 20 actions performed across all proposals.
- The removed propolsal is now possible to access through the global action list, so no additional methods to hadle it needed.

PR #219, opened 2025-03-16
feat: 214 add ProposalV1, last_touched_block_height

Tguntenaar:
WIP 

https://github.com/near-daos/sputnik-dao-contract/issues/214

### DIFF
--- a/sputnikdao2/src/bounties.rs
+++ b/sputnikdao2/src/bounties.rs
@@ -270,7 +270,12 @@ mod tests {
 
         assert_eq!(contract.get_last_proposal_id(), 2);
         assert_eq!(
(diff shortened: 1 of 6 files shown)


frol (commented):
@Tguntenaar Overall, this looks good to me. The only question I have is whether this is enough for the indexer

Tguntenaar:
> @Tguntenaar Overall, this looks good to me. The only question I have is whether this is enough for the indexer

It would not yet be sufficient. It wouldn't be backwards compatible since older proposals didn't store this field. Also the DAO can vote to remove proposals which then we lose track of what block height the proposal was removed and so also all previous `last_touched` references.

I will comeback to this, any suggestions?

frol:
Closing as it was taken over by #226

PR #234, opened 2026-03-24
tests: Fixed tests after global contracts support was deployed to mainnet

frol:


sputnikdao2/tests/test_bounties.rs
frol: Note: There are only `cargo fmt` changes in this file

